PR #1, opened 2024-02-20
X2Text adapter support added

gaya3-zipstack:
## What

Support added for X2Text adapters

## Why

To allow user to use X2Text adapters

## How

...

## Relevant Docs

-

## Related Issues or PRs

-

-

## Notes on Testing

...

## Screenshots

...

## Checklist


src/unstract/sdk/x2txt.py
```diff
@@ -0,0 +1,51 @@
...
+                # Add x2text service host, port and platform_service_key
+                x2text_metadata[
+                    X2TextConstants.X2TEXT_HOST
+                ] = self.tool.get_env_or_die(X2TextConstants.X2TEXT_HOST)
+                x2text_metadata[
+                    X2TextConstants.X2TEXT_PORT
+                ] = self.tool.get_env_or_die(X2TextConstants.X2TEXT_PORT)
+                x2text_metadata[
+                    X2TextConstants.PLATFORM_SERVICE_API_KEY
+                ] = self.tool.get_env_or_die(
```
chandrasekharan-zipstack, line 38: @gaya3-zipstack will this work fine for PromptStudio in a multitenant env as well since it runs from the backend?

pdm.lock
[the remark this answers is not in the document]
gaya3-zipstack: When i try pdm self update, I see an issue. How do you run it?
Installing collected packages: dep-logic, pdm
ERROR: Could not install packages due to an OSError: [Errno 13] Permission denied: '/usr/local/lib/python3.10/dist-packages/dep_logic-0.0.4.dist-info'
Consider using the `--user` option or check the permissions.

Hence I ve been with the old pdm till now

src/unstract/sdk/x2txt.py
gaya3-zipstack: @chandrasekharan-zipstack  If we are able to provide the platform service API key from the environment, then I don't see a problem.
PS: This is the same approach that is being followed for getting adapter config also from platform service.

chandrasekharan-zipstack: Ah now I remember - we follow [a hacky approach](https://github.com/Zipstack/Pandora/blob/b0b43085c9b881ab859691ccb12384d19412ffd8/backend/prompt_studio/prompt_studio_core/prompt_ide_base_tool.py#L33-L36) where we override this function and explicitly return the key.
I don't recommend this approach however it was a solution implemented for the sake of time

chandrasekharan-zipstack (approved):
LGTM

src/unstract/sdk/x2txt.py
```diff
@@ -0,0 +1,51 @@
...
+from unstract.sdk.constants import LogLevel
+from unstract.sdk.tool.base import BaseTool
+
+
+class X2Text(metaclass=ABCMeta):
+    def __init__(self, tool: BaseTool):
+        self.tool = tool
+        self.x2text_adapters = adapters
+
+    def get_x2text(self, adapter_instance_id: str) -> X2TextAdapter:
```
chandrasekharan-zipstack, line 18: Shouldn't this return type be `Optional[X2TextAdapter]`?

jaseemjaskp (approved):

PR #33, opened 2024-04-04
Fix handling of indexing error

hari-kuriakose:
## What

- Reuse `CustomException(str(original_err)) from original_err` pattern everywhere
- Add suggestive action for index URL **404 not found** error

## Why

For seamless E2E error propagation. We have already started this, need to follow everywhere.

## How

Since we use raise Exception from e pattern, the original exception traceback is preserved and pushed downstream.

## Related Issues or PRs

- https://github.com/Zipstack/unstract/pull/234

## Notes on Testing

Downstream is able to retrieve complete error trace.

## Screenshots

See `ERROR middleware.exception` in the logs:

```
{"type": "LOG", "stage": "TOOL_RUN", "level": "ERROR", "log": "Error adding nodes to vector db: Unexpected Response: 404 (Not Found)\nRaw responsecontent:\nb'404 page not found\\n'", "emitted_at": "2024-04-04T10:57:32.190425"}
[04/Apr/2024 10:57:32] ERROR middleware.exception: POST http://frontend.unstract.localhost/api/v1/unstract/mock_org/prompt-studio/index-document/ 400 IndexingAPIError('Index not found. Please check vector db adapter settings.')
Bad Request: /api/v1/unstract/mock_org/prompt-studio/index-document/
[04/Apr/2024 10:57:32] WARNING django.request: Bad Request: /api/v1/unstract/mock_org/prompt-studio/index-document/
172.19.0.3 - - [04/Apr/2024:10:57:32 +0000] "POST /api/v1/unstract/mock_org/prompt-studio/index-document/ HTTP/1.1" 400 132 "http://frontend.unstract.localhost/mock_org/tools/a9206e99-ae6c-45b3-80d2-c07b8d85c3f6" "Mozilla/5.0 (X11; Linux x86_64; rv:109.0) Gecko/20100101 Firefox/119.0"
```

## Checklist

### DIFF
--- a/src/unstract/sdk/exceptions.py
+++ b/src/unstract/sdk/exceptions.py
@@ -12,5 +12,6 @@ class SdkError(Exception):
 
 class IndexingError(SdkError):
     def __init__(self, message: str = ""):
-        prefix = "Error with indexing. "
-        super().__init__(prefix + message)
+        if "404" in message:
+            message = "Index not found. Please check vector db adapter settings."
+        super().__init__(message)
--- a/src/unstract/sdk/index.py
+++ b/src/unstract/sdk/index.py
@@ -293,7 +293,7 @@ class ToolIndex:
                     f"Error adding nodes to vector db: {e}",
                     level=LogLevel.ERROR,
                 )
-                raise SdkError(f"Error adding nodes to vector db: {e}")
+                raise IndexingError(str(e)) from e
             self.tool.stream_log("Added nodes to vector db")
 
         self.tool.stream_log("File has been indexed successfully")


nehabagdia (approved):

src/unstract/sdk/exceptions.py
```diff
@@ -12,5 +12,6 @@ def __str__(self) -> str:
 
 class IndexingError(SdkError):
     def __init__(self, message: str = ""):
-        prefix = "Error with indexing. "
-        super().__init__(prefix + message)
+        if "404" in message:
```
chandrasekharan-zipstack, line 15: Unsure if this is the best way to check - as the presence of a string `404` anywhere in the message will cause it to be overridden.

Since this is a 404 from some internal library, we could ideally have a separate `IndexNotFoundError` which can be caught by the `backend` and eventually return 404 error code itself

src/unstract/sdk/exceptions.py
```diff
@@ -12,5 +12,6 @@ def __str__(self) -> str:
 
 class IndexingError(SdkError):
     def __init__(self, message: str = ""):
-        prefix = "Error with indexing. "
-        super().__init__(prefix + message)
```
chandrasekharan-zipstack, line 16: @hari-kuriakose this prefix is needed as I use it to append context as we bubble up the message [here](https://github.com/Zipstack/unstract-sdk/blob/49635b1557f0abd72efea382611d5249b37a7ad6/src/unstract/sdk/index.py#L235).
image

Otherwise we wouldn't know where the error is coming from

chandrasekharan-zipstack (commented):
This would cause a regression in the error message itself, since `prefix` is removed. My suggestion is to create a separate error class for this scenario

src/unstract/sdk/exceptions.py
hari-kuriakose: @chandrasekharan-zipstack Most importantly we do not need a "Error with indexing" prefix when the type of exception itself is `IndexingError`.

src/unstract/sdk/exceptions.py
hari-kuriakose: @chandrasekharan-zipstack We do not need "Error with indexing" prefix when the type of exception itself is `IndexingError`. Downstream should handle this accordingly.

chandrasekharan-zipstack: @hari-kuriakose in that case, from the backend we would be prefixing `Error with indexing. ` for all `IndexingError` exceptions.
That would also imply that your particular message would be rendered as 
`Error with indexing. Index not found. Please check vector db adapter settings.`

I think this would work - we could review the error messages for cases which sound redundant (if any) later

src/unstract/sdk/exceptions.py
chandrasekharan-zipstack: @hari-kuriakose I'm okay with this assumption until we hit a blocker.
One benefit that we might miss out on by not explicitly differentiating `IndexingError` and `IndexNotFoundError` is that the BE can be clever enough to return `404` error code in case of the latter.
Either way, I think in the existing system we return 400 / 500 which needs to be looked into. Since it pertains to the `unstract` repo we can address it separately

chandrasekharan-zipstack (approved):

chandrasekharan-zipstack:
@hari-kuriakose how do we plan to release this? As a patch or do we plan to bundle it with the llama index upgrade?

hari-kuriakose:
> @hari-kuriakose how do we plan to release this? As a patch or do we plan to bundle it with the llama index upgrade?

@chandrasekharan-zipstack This is just a patch and hence not directly coupled with anything.
We can batch it with whichever release that seems suitable.

PR #36, opened 2024-04-16
Index and LLM fixes

Deepak-Kesavan:
## What

- Read file directly if it is ` txt`
- Updated regex to extract JSON
- Update `unstract-adapters` version

## Why

...

## How

...

## Relevant Docs

-

## Related Issues or PRs

-

-

## Notes on Testing

...

## Screenshots

...

## Checklist

### DIFF
--- a/src/unstract/sdk/index.py
+++ b/src/unstract/sdk/index.py
@@ -236,13 +236,17 @@ class ToolIndex:
         full_text = []
         extracted_text = ""
         try:
-            x2text = X2Text(tool=self.tool)
-            x2text_adapter_inst: X2TextAdapter = x2text.get_x2text(
-                adapter_instance_id=x2text_adapter
-            )
-            extracted_text = x2text_adapter_inst.process(
-                input_file_path=file_path, output_file_path=output_file_path
-            )
+            if not output_file_path:
+                with open(file_path, encoding="utf-8") as file:
+                    extracted_text = file.read()
+            else:
+                x2text = X2Text(tool=self.tool)
+                x2text_adapter_inst: X2TextAdapter = x2text.get_x2text(
+                    adapter_instance_id=x2text_adapter
+                )
+                extracted_text = x2text_adapter_inst.process(
+                    input_file_path=file_path, output_file_path=output_file_path
+                )
         except AdapterError as e:
             # Wrapping AdapterErrors with SdkError
             raise IndexingError(str(e)) from e
--- a/src/unstract/sdk/llm.py
+++ b/src/unstract/sdk/llm.py
@@ -22,7 +22,7 @@ logger = logging.getLogger(__name__)
 class ToolLLM:
     """Class to handle LLMs for Unstract Tools."""
 
-    code_block_regex = re.compile(r"```.*?\n(.*?)\n```", re.DOTALL)
+    json_regex = re.compile(r"\{(?:.|\n)*\}")
 
     def __init__(
         self,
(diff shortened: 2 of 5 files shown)


src/unstract/sdk/index.py
```diff
@@ -236,13 +236,17 @@ def index_file(
...
         extracted_text = ""
         try:
-            x2text = X2Text(tool=self.tool)
-            x2text_adapter_inst: X2TextAdapter = x2text.get_x2text(
-                adapter_instance_id=x2text_adapter
-            )
-            extracted_text = x2text_adapter_inst.process(
-                input_file_path=file_path, output_file_path=output_file_path
-            )
+            if not output_file_path:
```
gaya3-zipstack, line 239: Based on what does the caller understand whether output_file_path needs to be passed or not? @Deepak-Kesavan

gaya3-zipstack (approved):
Changes seem fine. Left one question for clarity.

src/unstract/sdk/llm.py
```diff
@@ -22,7 +22,7 @@
 class ToolLLM:
     """Class to handle LLMs for Unstract Tools."""
 
-    code_block_regex = re.compile(r"```.*?\n(.*?)\n```", re.DOTALL)
+    json_regex = re.compile(r"\{(?:.|\n)*\}")
```
chandrasekharan-zipstack, line 25: Can you update the PR description with why this change was needed?

src/unstract/sdk/index.py
```diff
@@ -236,13 +236,17 @@ def index_file(
...
+                with open(file_path, encoding="utf-8") as file:
+                    extracted_text = file.read()
```
chandrasekharan-zipstack, line 241: This may not work as expected if the file is a PDF right? I understand this is probably for the use case of quickly extracting text in the context of summaries but can you update the PR description as to why this change was done?

src/unstract/sdk/index.py
Deepak-Kesavan: @gaya3-zipstack Currently we use `output_file_path`, to store the extracted text. If the `output_file_path` is not present, the indexing is for `summarize` and the file is already a text.

`output_file_path` is optional and it is added in the doc string of the `index_file` function

src/unstract/sdk/index.py
Deepak-Kesavan: @chandrasekharan-zipstack There is no scenario as of now where the input file is a PDF and output path is not present.

nehabagdia (approved):

src/unstract/sdk/index.py
chandrasekharan-zipstack: My concern here is that, the function's behaviour needs to be explicit based on its arguments. The way we've currently handled - the caller needs to inspect the code to understand its behaviour. We could have ideally added a param `extract_text: bool` that can carry out the desired functionality of skipping a call to LLM whisperer.

By continuing to vaguely differentiate logic based on an unrelated param, we'd only find it increasingly difficult to maintain / add features

src/unstract/sdk/index.py
chandrasekharan-zipstack: @Deepak-Kesavan 
> If the output_file_path is not present, the indexing is for summarize and the file is already a text.

This won't be immediately obvious to a developer, I would advice against implicitly performing such logic. Anyway this function is due for a refactor

src/unstract/sdk/llm.py
chandrasekharan-zipstack: If I understand it right, we are ignore all JSONs returned by the LLM calls right? Do review this, but this might imply a regression for example in the context of workflow generation with a prompt (I know we barely use this feature but still)

PR #40, opened 2024-04-22
Added adapter_instance_id and run_id

Deepak-Kesavan:
## What

- Updated SDK to push `adapter_instance_id` and `run_id`
- Reset token count after each usage push
- SDK version bump

## Why

...

## How

...

## Relevant Docs

-

## Related Issues or PRs

-

-

## Notes on Testing

...

## Screenshots

...

## Checklist

### DIFF
--- a/src/unstract/sdk/audit.py
+++ b/src/unstract/sdk/audit.py
@@ -15,15 +15,6 @@ class Audit(StreamMixin):
 
     Attributes:
         None
-
-    Example usage:
-        audit = Audit()
-        audit.push_usage_data(
-            token_counter,
-            workflow_id,
-            execution_id,
-            external_service,
-            event_type)
     """
 
     def __init__(self, log_level: LogLevel = LogLevel.INFO) -> None:
@@ -33,7 +24,7 @@ class Audit(StreamMixin):
         self,
         platform_api_key: str,
         token_counter: TokenCountingHandler = None,
-        external_service: str = "",
+        model_name: str = "",
         event_type: CBEventType = None,
         **kwargs,
     ) -> None:
@@ -43,7 +34,7 @@ class Audit(StreamMixin):
             platform_api_key (str): The platform API key.
             token_counter (TokenCountingHandler, optional): The token counter
                 object. Defaults to None.
-            external_service (str, optional): The name of the external service.
+            model_name (str, optional): The name of the model.
                 Defaults to "".
             event_type (CBEventType, optional): The type of the event. Defaults
                 to None.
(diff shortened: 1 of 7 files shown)


johnyrahul (approved):

src/unstract/sdk/audit.py
```diff
@@ -33,23 +24,26 @@ def push_usage_data(
...
         Args:
+            platform_api_key (str): The platform API key.
             token_counter (TokenCountingHandler, optional): The token counter
-              object. Defaults to None.
+                object. Defaults to None.
+            model_name (str, optional): The name of the model.
+                Defaults to "".
+            event_type (CBEventType, optional): The type of the event. Defaults
+                to None.
             workflow_id (str, optional): The ID of the workflow. Defaults to "".
```
gaya3-zipstack, line 41: Do you want to list these args here? Ideally the args is just kwargs. You can list the expected kwargs. Explicitly mentioning these as args gives a picture as it they are unpacked here, which is not the case.

gaya3-zipstack (commented):
Needs adapter version roll. Breaking change as it has changes in set_callback_manager API.

src/unstract/sdk/audit.py
```diff
@@ -33,23 +24,26 @@ def push_usage_data(
...
             execution_id (str, optional): The ID of the execution. Defaults
-              to "".
-            external_service (str, optional): The name of the external service.
-              Defaults to "".
-            event_type (CBEventType, optional): The type of the event. Defaults
-              to None.
+                to "".
+            adapter_instance_id (str, optional): The adapter instance ID.
+                Defaults to "".
+            run_id (str, optional): The run ID. Defaults to "".
```
chandrasekharan-zipstack, line 46: NIT: Update kwargs as a separate section and retain only actually args here

src/unstract/sdk/audit.py
Deepak-Kesavan: @gaya3-zipstack I added it so that we know that these args are sent just by looking into the doc string.

src/unstract/sdk/audit.py
```diff
@@ -84,6 +85,7 @@ def push_usage_data(
             response = requests.post(
                 url, headers=headers, json=data, timeout=30
             )
+            token_counter.reset_counts()
```
chandrasekharan-zipstack, line 88: NIT: I'm assuming this resets the token usage for the counter - it might be good to place this in a `finally` block to execute irrespective of whether an error occurred or not

src/unstract/sdk/audit.py
gaya3-zipstack: I agree. All I am saying is that we could do something like
Args:
            kwargs:
                Expected kwargs:
                    - adapter_instance_id
                    - workflow_id
                    - ....

chandrasekharan-zipstack (approved):
LGTM - added some NIT comments, please try to incorporate.
Make sure the version is bumped appropriately

src/unstract/sdk/audit.py
Deepak-Kesavan: Done

src/unstract/sdk/audit.py
Deepak-Kesavan: Done

src/unstract/sdk/audit.py
Deepak-Kesavan: Done

gaya3-zipstack (approved):
Looks good

PR #52, opened 2024-05-23
[FIX] Exception enhancements for Prompt Service

harini-venkataraman:
## What
Exception enhancements for Prompt service.
...

## Why
PR surfaces proper error messages to frontend.
...

## How
Handling specific error messages.
...

## Relevant Docs
https://requests.readthedocs.io/en/latest/user/quickstart/#errors-and-exceptions

...
## Related Issues or PRs

...

NA
...

## Notes on Testing
Try running the prompt with appropriate erroneous situations. 
...

## Screenshots

image

image

image

image

...

## Checklist

### DIFF
--- a/src/unstract/sdk/prompt.py
+++ b/src/unstract/sdk/prompt.py
@@ -2,13 +2,7 @@ import logging
 from typing import Any, Optional
 
 import requests
-from requests import (
-    ConnectionError,
-    RequestException,
-    Response,
-    Timeout,
-    TooManyRedirects,
-)
+from requests import ConnectionError, RequestException, Response, Timeout
 
 from unstract.sdk.constants import LogLevel, PromptStudioKeys, ToolEnv
 from unstract.sdk.helper import SdkHelper
@@ -81,14 +75,11 @@ class PromptTool:
             self._stringify_and_stream_err(connect_err, msg)
             result["error"] = msg
         except Timeout as time_out:
-            msg = """Request to run prompt has timed out.
-                   Probable causes would be connectivity issues in LLMs."""
+            msg = (
+                "Request to run prompt has timed out. "
+                "Probable causes would be connectivity issues in LLMs."
+            )
             self._stringify_and_stream_err(time_out, msg)
-            result["error"] = msg
-        except TooManyRedirects as too_many_redirects:
-            msg = """Too many redirects while connecting to prompt service."
-                    Try clearing cache or contact admin"""
-            self._stringify_and_stream_err(too_many_redirects, msg)
             result["error"] = msg
         except RequestException as e:
             # Extract error information from the response if available
--- a/src/unstract/sdk/prompt.py
+++ b/src/unstract/sdk/prompt.py
@@ -1,7 +1,14 @@
 from typing import Any, Optional
 
 import requests
-from requests import RequestException, Response
+from requests import (
+    ConnectionError,
+    HTTPError,
+    RequestException,
+    Response,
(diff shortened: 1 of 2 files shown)


src/unstract/sdk/prompt.py
```diff
@@ -63,17 +68,28 @@ def _post_call(
             "structure_output": "",
         }
         url: str = f"{self.base_url}/{url_path}"
-        headers: dict[str, str] = {
-            "Authorization": f"Bearer {self.bearer_token}"
-        }
+        headers: dict[str, str] = {"Authorization": f"Bearer {self.bearer_token}"}
+        response: Response = Response()
```
chandrasekharan-zipstack, line 72: NIT: This seems valid since we've explicitly handled `Timeout` and `ConnectionError` but please keep in mind that if the `response` is not actually returned in line 75 and there's a `RequestException` while calling `requests.post()` we might end up making the below calls which could result in an erroneous behaviour
```
response.headers.get("Content-Type", "").lower()
response.json()
response.text
```
I'm not sure if these might return anything valid for the default constructed `Response` object

chandrasekharan-zipstack (approved):
Looks good for the most part - left some minor comments. Please consider the suggestions if they seem valid

src/unstract/sdk/prompt.py
```diff
@@ -91,6 +107,12 @@ def _post_call(
             )
         return result
```
gaya3-zipstack, line 108: Since we are returning result here, will there be a case where we could get "Used before defined" error for the variable result when we end up in any unhandled exception here? Should we add an additional handler to catch Exception in general too? @harini-venkataraman

src/unstract/sdk/prompt.py
harini-venkataraman: @chandrasekharan-zipstack 
1. We have handled `Timeout` and `ConnectionError` - which does not have the response variable. So this doesn't go into RequestException
2. If RequestException doesnt return any Response object - [This validation](https://github.com/Zipstack/unstract-sdk/blob/b2bb0786d28ea500f6fba5bceebcbf0626d74662/src/unstract/sdk/prompt.py#L101) takes care and the stringified error message will be returned.

src/unstract/sdk/prompt.py
harini-venkataraman: @gaya3-zipstack This will be handled by the variable initialized [here](https://github.com/Zipstack/unstract-sdk/blob/b2bb0786d28ea500f6fba5bceebcbf0626d74662/src/unstract/sdk/prompt.py#L64). So, this error will not be thrown in this case.

src/unstract/sdk/prompt.py
```diff
@@ -63,17 +68,28 @@ def _post_call(
...
             response.raise_for_status()
             result["status"] = "OK"
             result["structure_output"] = response.text
+        except ConnectionError as connect_err:
+            msg = "Unable to connect to prompt service. Please contact admin."
+            self._stringify_and_stream_err(connect_err, msg)
+            result["error"] = msg
+        except Timeout as time_out:
+            msg = """Request to run prompt has timed out.
+                   Probable causes would be connectivity issues in LLMs."""
```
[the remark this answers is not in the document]
harini-venkataraman: Addressed. Removed the multi line string and added as per suggestion.

gaya3-zipstack (approved):
Looks fine

chandrasekharan-zipstack (approved):

PR #55, opened 2024-05-30
SDK backward compatibility changes

gaya3-zipstack:
## What

- SDK version 0.28.0 was not backward compatible. This PR includes changes to support backward compatibility
-  integrated close() API for vector DB connection closures

## Why

- Being OSS we need backward compatibility to keep the community developers productive.

## How

- Renamed new functions
- Kept old functions intact to support backward compatibility
- Added deprecation warnings appropriately

## Relevant Docs

- NA

## Related Issues or PRs

- NA

- NA

## Notes on Testing

...

## Screenshots

...

## Checklist

### DIFF
--- a/src/unstract/sdk/embedding.py
+++ b/src/unstract/sdk/embedding.py
@@ -22,7 +22,7 @@ class Embedding:
         self,
         tool: BaseTool,
         adapter_instance_id: str = None,
-        usage_kwargs: dict[Any, Any] = None,
+        usage_kwargs: dict[Any, Any] = {},
     ):
         self._tool = tool
         self._adapter_instance_id = adapter_instance_id
--- a/src/unstract/sdk/embedding.py
+++ b/src/unstract/sdk/embedding.py
@@ -1,4 +1,4 @@
-from typing import Any
+from typing import Any, Optional
 
 from llama_index.core.base.embeddings.base import Embedding
 from llama_index.core.embeddings import BaseEmbedding
@@ -21,7 +21,7 @@ class Embedding:
     def __init__(
         self,
         tool: BaseTool,
-        adapter_instance_id: str = None,
+        adapter_instance_id: Optional[str] = None,
         usage_kwargs: dict[Any, Any] = {},
     ):
         self._tool = tool
--- a/src/unstract/sdk/embedding.py
+++ b/src/unstract/sdk/embedding.py
@@ -21,22 +21,27 @@ class Embedding:
     def __init__(
         self,
         tool: BaseTool,
-        adapter_instance_id: str,
+        adapter_instance_id: str = None,
         usage_kwargs: dict[Any, Any] = None,
     ):
         self._tool = tool
         self._adapter_instance_id = adapter_instance_id
-        self._embedding_instance: BaseEmbedding = self._get_embedding()
-        self._length: int = self._get_embedding_length()
+        self._embedding_instance: BaseEmbedding = None
+        self._length: int = None
+        self._usage_kwargs = usage_kwargs if usage_kwargs else dict()
+        self._initialise()
 
-        self._usage_kwargs = usage_kwargs.copy()
-        self._usage_kwargs["adapter_instance_id"] = adapter_instance_id
-        platform_api_key = self._tool.get_env_or_die(ToolEnv.PLATFORM_API_KEY)
-        CallbackManager.set_callback_manager(
-            platform_api_key=platform_api_key,
-            model=self._embedding_instance,
-            kwargs=self._usage_kwargs,
-        )
+    def _initialise(self):
+        if self._adapter_instance_id:
+            self._embedding_instance = self._get_embedding()
+            self._length: int = self._get_embedding_length()
+            self._usage_kwargs["adapter_instance_id"] = self._adapter_instance_id
+            platform_api_key = self._tool.get_env_or_die(ToolEnv.PLATFORM_API_KEY)
+            CallbackManager.set_callback(
+                platform_api_key=platform_api_key,
+                model=self._embedding_instance,
+                kwargs=self._usage_kwargs,
+            )
 
     def _get_embedding(self) -> BaseEmbedding:
         """Gets an instance of LlamaIndex's embedding object.
@@ -48,6 +53,10 @@ class Embedding:
             BaseEmbedding: Embedding instance
         """
         try:
+            if not self._adapter_instance_id:
+                raise EmbeddingError(
+                    "Adapter instance ID not set. " "Initialisation failed"
+                )
             embedding_config_data = ToolAdapter.get_adapter_config(
                 self._tool, self._adapter_instance_id
             )
@@ -79,9 +88,18 @@ class Embedding:
         embedding_dimension = len(embedding_list)
         return embedding_dimension
 
-    @deprecated("Use the new class Embedding")
+    @deprecated("Deprecated class and method. Use Embedding instead of ToolEmbedding")
     def get_embedding_length(self, embedding: BaseEmbedding) -> int:
-        return self._get_embedding_length(embedding)
+        return self._get_embedding_length()
+
+    @deprecated(
+        "Deprecated class and method. " "Use Embedding instead of ToolEmbedding"
+    )
+    def get_embedding(self, adapter_instance_id: str) -> BaseEmbedding:
+        if not self._embedding_instance:
+            self._adapter_instance_id = adapter_instance_id
+            self._initialise()
+        return self._embedding_instance
 
 
 # Legacy
--- a/src/unstract/sdk/embedding.py
+++ b/src/unstract/sdk/embedding.py
@@ -28,7 +28,7 @@ class Embedding:
         self._adapter_instance_id = adapter_instance_id
         self._embedding_instance: BaseEmbedding = None
         self._length: int = None
-        self._usage_kwargs = usage_kwargs if usage_kwargs else dict()
+        self._usage_kwargs = usage_kwargs
         self._initialise()
 
     def _initialise(self):
@@ -99,13 +99,11 @@ class Embedding:
         """
         return self._embedding_instance.class_name()
 
-    @deprecated("Deprecated class and method. Use Embedding instead of ToolEmbedding")
+    @deprecated("Use Embedding instead of ToolEmbedding")
     def get_embedding_length(self, embedding: BaseEmbedding) -> int:
         return self._get_embedding_length()
 
-    @deprecated(
-        "Deprecated class and method. " "Use Embedding instead of ToolEmbedding"
-    )
+    @deprecated("Use Embedding instead of ToolEmbedding")
     def get_embedding(self, adapter_instance_id: str) -> BaseEmbedding:
         if not self._embedding_instance:
             self._adapter_instance_id = adapter_instance_id
--- a/src/unstract/sdk/embedding.py
+++ b/src/unstract/sdk/embedding.py
@@ -88,6 +88,17 @@ class Embedding:
         embedding_dimension = len(embedding_list)
         return embedding_dimension
 
+    def get_class_name(self) -> str:
+        """Gets the class name of the Llama Index Embedding.
+
+        Args:
+            NA
+
+            Returns:
+                Class name
+        """
+        return self._embedding_instance.class_name()
+
     @deprecated("Deprecated class and method. Use Embedding instead of ToolEmbedding")
     def get_embedding_length(self, embedding: BaseEmbedding) -> int:
         return self._get_embedding_length()
--- a/src/unstract/sdk/index.py
+++ b/src/unstract/sdk/index.py
@@ -28,7 +28,7 @@ class Index:
         # TODO: Inherit from StreamMixin and avoid using BaseTool
         self.tool = tool
 
-    def query_text_from_index(
+    def query_index(
         self,
         embedding_instance_id: str,
         vector_db_instance_id: str,
@@ -373,7 +373,7 @@ class Index:
         hashed_index_key = ToolUtils.hash_str(json.dumps(index_key, sort_keys=True))
         return hashed_index_key
 
-    @deprecated("Deprecated class and method. Use Index and index() instead")
+    @deprecated("Instantiate Index and call index() instead")
     def index_file(
         self,
         tool_id: str,
@@ -406,7 +406,7 @@ class Index:
     def get_text_from_index(
         self, embedding_type: str, vector_db: str, doc_id: str
     ) -> Optional[str]:
-        return self.query_text_from_index(
+        return self.query_index(
             embedding_instance_id=embedding_type,
             vector_db_instance_id=vector_db,
             doc_id=doc_id,
--- a/src/unstract/sdk/index.py
+++ b/src/unstract/sdk/index.py
@@ -10,6 +10,7 @@ from llama_index.core.vector_stores import (
     VectorStoreQuery,
     VectorStoreQueryResult,
 )
+from typing_extensions import deprecated
 from unstract.adapters.exceptions import AdapterError
 
 from unstract.sdk.adapters import ToolAdapter
@@ -27,7 +28,7 @@ class Index:
         # TODO: Inherit from StreamMixin and avoid using BaseTool
         self.tool = tool
 
-    def get_text_from_index(
+    def query_text_from_index(
         self,
         embedding_instance_id: str,
         vector_db_instance_id: str,
@@ -114,7 +115,7 @@ class Index:
         )
         return "\n".join(new_context_lines)
 
-    def index_file(
+    def index(
         self,
         tool_id: str,
         embedding_instance_id: str,
@@ -150,7 +151,7 @@ class Index:
         Returns:
             str: A unique ID for the file and indexing arguments combination
         """
-        doc_id = self.generate_file_id(
+        doc_id = self._generate_file_id(
             tool_id=tool_id,
             vector_db=vector_db_instance_id,
             embedding=embedding_instance_id,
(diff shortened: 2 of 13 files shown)


src/unstract/sdk/index.py
```diff
@@ -186,123 +187,138 @@ def index_file(
...
-            n: VectorStoreQueryResult = vector_db.query(query=q)
-            if len(n.nodes) > 0:
-                doc_id_found = True
-                self.tool.stream_log(f"Found {len(n.nodes)} nodes for {doc_id}")
-            else:
-                self.tool.stream_log(f"No nodes found for {doc_id}")
-        except Exception as e:
-            self.tool.stream_log(
-                f"Error querying {vector_db_instance_id}: {e}", level=LogLevel.ERROR
+            # Checking if document is already indexed against doc_id
```
gaya3-zipstack, line 191: The changes that are showing up here is due to the added try-finally block to close the vector db connection. Mainly white space indentation. Core-logic is the same.

src/unstract/sdk/x2txt.py
```diff
@@ -59,8 +68,16 @@ def process(
...
     ) -> str:
-        return self.x2text_instance.process(input_file_path, output_file_path, **kwargs)
+        return self._x2text_instance.process(
+            input_file_path, output_file_path, **kwargs
+        )
 
-    @deprecated("Use the class instance")
+    @deprecated(
+        "Deprecated method. Use the latest X2Text to "
+        "instantiate and then invoke process()"
```
hari-kuriakose, line 77: @gaya3-zipstack I believe `deprecated` library will throw warning that it is deprecated, we do not need to mention again.

In that case let's change the message to `Use process() method of X2Text instance.`.

src/unstract/sdk/vector_db.py
```diff
@@ -152,18 +165,31 @@ def add(
...
+
+        Args:
+            NA
+
+            Returns:
+                Class name
+        """
+        return self._vector_db_instance.class_name()
+
+    @deprecated("Deprecated class and method. Use VectorDB instead of ToolVectorDB")
```
hari-kuriakose, line 184: @gaya3-zipstack `deprecated` library will throw warning that it is deprecated, we do not need to mention again.

In that case, let's change message to `Use VectorDB instead of ToolVectorDB.`.

src/unstract/sdk/vector_db.py
```diff
@@ -30,24 +30,33 @@ class VectorDB:
 
     vector_db_adapters = adapters
     DEFAULT_EMBEDDING_DIMENSION = 1536
+    EMBEDDING_INSTANCE_ERROR = (
+        "Vector DB does not have an embedding initialised."
+        "Migrate to VectorDB instead of deprecated ToolVectorDB "
+        "to use the latest features"
```
hari-kuriakose, line 36: @gaya3-zipstack Let's simply say:
```
VectorDB does not have en embedding initialised.
ToolVectorDB is deprecated. Migrate to VectorDB instead.
```

BTW, if we are throwing an error here and not allowing to proceed, then this is not actually backward compatible.
But should be relatively ok for now, let's see.

src/unstract/sdk/utils/callback_manager.py
```diff
@@ -130,3 +131,22 @@ def get_tokenizer(
         except OSError as e:
             logger.warning(str(e))
             return fallback_tokenizer
+
+    @staticmethod
+    @deprecated("Deprecated method. Use the latest set_callback instead")
```
hari-kuriakose, line 136: @gaya3-zipstack `deprecated` library will throw warning that it is deprecated, we do not need to mention again.

In that case let's change message to `Use set_callback() instead.`.

src/unstract/sdk/ocr.py
```diff
@@ -8,47 +8,61 @@
...
     def process(
         self, input_file_path: str, output_file_path: Optional[str] = None
     ) -> str:
-        return self.ocr_instance.process(input_file_path, output_file_path)
+        return self._ocr_instance.process(input_file_path, output_file_path)
 
-    @deprecated("Use the class instance")
+    @deprecated(
+        "Deprecated method. Use the latest OCR to instantiate "
+        "and then invoke process()"
```
hari-kuriakose, line 62: @gaya3-zipstack `deprecated` library will throw warning that it is deprecated, we do not need to mention again.

In that case let's change message to `Use process() method of OCR instance.`.

src/unstract/sdk/llm.py
```diff
@@ -120,10 +127,75 @@ def _get_max_tokens(self, reserved_for_output: int = 0) -> int:
...
+
+    @deprecated("Deprecated class and method. Use LLM instead of ToolLLM")
     def get_llm(self, adapter_instance_id: Optional[str] = None) -> LlamaIndexLLM:
+        if not self._llm_instance:
+            self._adapter_instance_id = adapter_instance_id
+            self._initialise()
         return self._llm_instance
 
+    @classmethod
+    @deprecated("Deprecated class and method. Use LLM and complete() instead")
```
hari-kuriakose, line 160: @gaya3-zipstack `deprecated` library will throw warning that it is deprecated, we do not need to mention again.

In that case let's change message to `Use complete() method of LLM instance instead.`.

src/unstract/sdk/llm.py
```diff
@@ -120,10 +127,75 @@ def _get_max_tokens(self, reserved_for_output: int = 0) -> int:
...
+
+        Args:
+            NA
+
+            Returns:
+                Class name
+        """
+        return self._llm_instance.class_name()
+
+    @deprecated("Deprecated class and method. Use LLM instead of ToolLLM")
```
hari-kuriakose, line 152: @gaya3-zipstack `deprecated` library will throw warning that it is deprecated, we do not need to mention again.

In that case let's change message simply to `Use LLM instead of ToolLLM.`.

src/unstract/sdk/index.py
```diff
@@ -357,6 +373,45 @@ def generate_file_id(
...
+            file_path=file_path,
+            chunk_size=chunk_size,
+            chunk_overlap=chunk_overlap,
+            reindex=reindex,
+            file_hash=file_hash,
+            output_file_path=output_file_path,
+        )
+
+    @deprecated(
+        "Deprecated class and method. Use Index and query_texy_from_index() instead"
```
hari-kuriakose, line 404: @gaya3-zipstack `deprecated` library will throw warning that it is deprecated, we do not need to mention again.

In that case let's change message to `Use query_index() method of Index instance instead.`.

src/unstract/sdk/index.py
```diff
@@ -357,6 +373,45 @@ def generate_file_id(
         hashed_index_key = ToolUtils.hash_str(json.dumps(index_key, sort_keys=True))
         return hashed_index_key
 
+    @deprecated("Deprecated class and method. Use Index and index() instead")
```
hari-kuriakose, line 376: @gaya3-zipstack `deprecated` library will throw warning that it is deprecated, we do not need to mention again.

In that case let's change message to `Use index() method of Index instance instead.`.

src/unstract/sdk/index.py
```diff
@@ -27,7 +28,7 @@ def __init__(self, tool: BaseTool):
         # TODO: Inherit from StreamMixin and avoid using BaseTool
         self.tool = tool
 
-    def get_text_from_index(
+    def query_text_from_index(
```
hari-kuriakose, line 31: @gaya3-zipstack Can we rename this to simply `query_index()`?

src/unstract/sdk/embedding.py
```diff
@@ -79,9 +88,29 @@ def _get_embedding_length(self) -> int:
...
+
+        Args:
+            NA
+
+            Returns:
+                Class name
+        """
+        return self._embedding_instance.class_name()
+
+    @deprecated("Deprecated class and method. Use Embedding instead of ToolEmbedding")
```
hari-kuriakose, line 102: @gaya3-zipstack `deprecated` library will throw warning that it is deprecated, we do not need to mention again.

In that case let's change message to `Use Embedding instead of ToolEmbedding.`.

src/unstract/sdk/embedding.py
```diff
@@ -79,9 +88,29 @@ def _get_embedding_length(self) -> int:
...
     def get_embedding_length(self, embedding: BaseEmbedding) -> int:
-        return self._get_embedding_length(embedding)
+        return self._get_embedding_length()
+
+    @deprecated(
+        "Deprecated class and method. " "Use Embedding instead of ToolEmbedding"
```
hari-kuriakose, line 107: @gaya3-zipstack `deprecated` library will throw warning that it is deprecated, we do not need to mention again.

In that case let's change message to `Use Embedding instead of ToolEmbedding.`.

src/unstract/sdk/index.py
```diff
@@ -186,123 +187,138 @@ def index_file(
...
+                    # TODO: Phase 2:
+                    # Post insertion to VDB, use query using doc_id and
+                    # store all the VDB ids to a table against the doc_id
+                    # During deletion for cases where metadata filtering
+                    # does not work, these ids can be used for direct deletion
+                    # This new table will also act like an audit trail for
+                    # all nodes that were added to the VDB by Unstract
+                    # Once this is in place, the overridden implementation
+                    # of prefixing ids with doc_id before adding to VDB
+                    # can be removed
```
hari-kuriakose, line 302: @gaya3-zipstack Is the reason why we need to do this because the vector store query may fail to return relevant results?

src/unstract/sdk/llm.py
```diff
@@ -45,18 +46,22 @@ def __init__(
...
-
-        self._usage_kwargs = usage_kwargs.copy()
-        self._usage_kwargs["adapter_instance_id"] = adapter_instance_id
-        platform_api_key = self._tool.get_env_or_die(ToolEnv.PLATFORM_API_KEY)
-        CallbackManager.set_callback_manager(
-            platform_api_key=platform_api_key,
-            model=self._llm_instance,
-            kwargs=self._usage_kwargs,
-        )
+        self._llm_instance: LlamaIndexLLM = None
```
hari-kuriakose, line 49: @gaya3-zipstack If I want to use the LlamaIndex LLM instance directly for some reason, then is accessing like `llm._llm_instance` enough to do this from hereon?

src/unstract/sdk/llm.py
hari-kuriakose, line 1: @gaya3-zipstack Docstring for LLM class can instead say something like:
`Interface to handle all LLM interactions.`

We need to gradually move away from SDK from being meant only for our tools.

hari-kuriakose (approved):
@gaya3-zipstack LGTM overall. Great progress on backward compatibility.

Added some comments though.

src/unstract/sdk/x2txt.py
gaya3-zipstack: Done

src/unstract/sdk/vector_db.py
gaya3-zipstack: Done

src/unstract/sdk/vector_db.py
gaya3-zipstack: This will not be faced by people who use the old SDK. This will be faced by people who call this function without initialising an embedding instance. ANyway, I think the error message was not very helpful. I have changed it now.

src/unstract/sdk/utils/callback_manager.py
gaya3-zipstack: Done

src/unstract/sdk/ocr.py
gaya3-zipstack: Done

src/unstract/sdk/llm.py
gaya3-zipstack: Done

src/unstract/sdk/llm.py
gaya3-zipstack: Done

src/unstract/sdk/index.py
gaya3-zipstack: Done

src/unstract/sdk/embedding.py
gaya3-zipstack: Done

src/unstract/sdk/index.py
gaya3-zipstack: This is the proposed design  delete vector db records while indexing/re-indexing so that we can handle cases like Pinecone where metadata filtering does not work.

gaya3-zipstack: https://zipstack.atlassian.net/browse/UN-1352

src/unstract/sdk/llm.py
gaya3-zipstack: Yes. But _llm_instance is a protected field right? Should I add a accessor for that? I would prefer not to. But Python will anyway allow access to the variable I guess!

src/unstract/sdk/ocr.py
```diff
@@ -8,47 +8,58 @@
...
+from unstract.sdk.exceptions import OCRError
 from unstract.sdk.tool.base import BaseTool
 
 
 class OCR(metaclass=ABCMeta):
     def __init__(
         self,
         tool: BaseTool,
-        adapter_instance_id: str,
+        adapter_instance_id: str = None,
```
chandrasekharan-zipstack, line 19: @gaya3-zipstack please correct this type to `Optional[str]` (at all other places too)

src/unstract/sdk/vector_db.py
```diff
@@ -70,6 +79,10 @@ def _get_vector_db(self) -> Union[BasePydanticVectorStore, VectorStore]:
             Union[BasePydanticVectorStore, VectorStore]: Vector store instance
         """
         try:
+            if not self._adapter_instance_id:
+                raise VectorDBError(
+                    "Adapter instance ID not set. " "Initialisation failed"
```
chandrasekharan-zipstack, line 84:
```suggestion
                    "Adapter instance ID not set. Initialisation failed"
```

src/unstract/sdk/llm.py
```diff
@@ -120,10 +127,75 @@ def _get_max_tokens(self, reserved_for_output: int = 0) -> int:
...
+            new_kwargs.pop(key, None)
+
+        try:
+            response: CompletionResponse = llm.complete(prompt, **new_kwargs)
+            match = cls.json_regex.search(response.text)
+            if match:
+                response.text = match.group(0)
+            return {"response": response}
+        # TODO: Handle for all LLM providers
+        except OpenAIAPIError as e:
```
harini-venkataraman, line 190: @gaya3-zipstack Looks like this error is specific to OpenAI. can this be handled in the adapter level?

src/unstract/sdk/llm.py
```diff
@@ -120,10 +127,75 @@ def _get_max_tokens(self, reserved_for_output: int = 0) -> int:
...
+            msg = "OpenAI error: "
+            msg += e.message
+            if hasattr(e, "body") and "message" in e.body:
+                msg += e.body["message"]
+            if isinstance(e, OpenAIRateLimitError):
+                raise RateLimitError(msg)
+            raise LLMError(msg) from e
```
harini-venkataraman, line 197: @gaya3-zipstack Should `e` be surfaced to the end user?

src/unstract/sdk/llm.py
chandrasekharan-zipstack: Yes @harini-venkataraman - it should ideally be at the adapter level. I had to add this handling here when we still had some of the old changes. If this error is handled at the adapter level code for an LLM call we can remove this block @gaya3-zipstack

src/unstract/sdk/llm.py
chandrasekharan-zipstack: @harini-venkataraman over here, we take the user friendly message alone from the exception and pass it. It should ideally get prefixed with context on where the error is coming from - this is handled already I believe. 
The `raise exception from e` will only preserve the traceback for us and will not pass the exception itself to the user

src/unstract/sdk/llm.py
hari-kuriakose: @gaya3-zipstack Yes let's keep it protected, because normally it need not be accessed directly as you mentioned.

However there could be some Enterprise workflows (e.g. multi-doc chat) or a new LlamaIndex feature that we might need on a use case basis for which this would be required, that's all.

src/unstract/sdk/index.py
hari-kuriakose: On a general note, we might need to refrain from doing this and trying to move away from it in the long run, as it would be better not to fiddle with LlamaIndex internal bookkeeping at all, and instead focus on various indexing and retrieval strategies offered by the framework.

Ideally we should also aggressively flag any LlamaIndex issues found for a particular provider, and maybe even raise patches of our own.

src/unstract/sdk/index.py
hari-kuriakose: Done

src/unstract/sdk/llm.py
```diff
@@ -120,10 +127,75 @@ def _get_max_tokens(self, reserved_for_output: int = 0) -> int:
...
+    def get_class_name(self) -> str:
+        """Gets the class name of the Llama Index LLM.
+
+        Args:
+            NA
+
+            Returns:
+                Class name
+        """
+        return self._llm_instance.class_name()
```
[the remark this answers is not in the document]
gaya3-zipstack: This is being used to detect tokenizer. Will have to revisit that part. May remove it if not required then. This is a wrapper on llama-index equivalent. Hence retaining the same name for now.

src/unstract/sdk/llm.py
```diff
@@ -21,16 +21,17 @@
 
 
 class LLM:
-    """Class to handle LLMs for Unstract Tools."""
+    """Em"""
```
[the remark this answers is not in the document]
gaya3-zipstack: Done

src/unstract/sdk/ocr.py
gaya3-zipstack: Thank you!! Done.

src/unstract/sdk/llm.py
gaya3-zipstack: @harini-venkataraman @chandrasekharan-zipstack In order to take it to adapter level, we need to spport all these in adapter leve. Will reach there at some point ...

nehabagdia (approved):

src/unstract/sdk/llm.py
chandrasekharan-zipstack: Okay - until its done, let this remain until then

chandrasekharan-zipstack (approved):

PR #64, opened 2024-06-21
Fix/adding whisper hash in metadata

johnyrahul:
## What
- [x]  Need to update the PDM lock with changes in adapter version as  it is breaking change https://github.com/Zipstack/unstract-adapters/pull/53
Adding whisper has to meta_data.json

## Why

Whisper has can be used by upstream clients to make use of  LLM Whisperer'd highlight capability.

## How

 While performing the text extraction save whisper hash in meta_data.json if the text adapter instance used is LLM whisperer.

## Relevant Docs

-

## Related Issues or PRs

-

-

## Notes on Testing

...

## Screenshots

...

## Checklist

### DIFF
--- a/src/unstract/sdk/index.py
+++ b/src/unstract/sdk/index.py
@@ -127,6 +127,7 @@ class Index:
         reindex: bool = False,
         file_hash: Optional[str] = None,
         output_file_path: Optional[str] = None,
+        enable_highlight: bool = False,
         usage_kwargs: dict[Any, Any] = {},
     ) -> str:
         """Indexes an individual file using the passed arguments.
@@ -243,7 +244,9 @@ class Index:
                         tool=self.tool, adapter_instance_id=x2text_instance_id
                     )
                     extracted_text = x2text.process(
-                        input_file_path=file_path, output_file_path=output_file_path
+                        input_file_path=file_path,
+                        output_file_path=output_file_path,
+                        enable_highlight=enable_highlight,
                     )
             except AdapterError as e:
                 # Wrapping AdapterErrors with SdkError
@@ -400,9 +403,7 @@ class Index:
             output_file_path=output_file_path,
         )
 
-    @deprecated(
-        "Deprecated class and method. Use Index and query_index() instead"
-    )
+    @deprecated("Deprecated class and method. Use Index and query_index() instead")
     def get_text_from_index(
         self, embedding_type: str, vector_db: str, doc_id: str
     ) -> Optional[str]:
--- a/src/unstract/sdk/index.py
+++ b/src/unstract/sdk/index.py
@@ -23,6 +23,10 @@ from unstract.sdk.vector_db import VectorDB
 from unstract.sdk.x2txt import X2Text
 
 
+class Constants:
+    TOP_K = 5
+
+
 class Index:
     def __init__(self, tool: BaseTool):
         # TODO: Inherit from StreamMixin and avoid using BaseTool
@@ -73,7 +77,7 @@ class Index:
                     query_embedding=embedding.get_query_embedding(" "),
                     doc_ids=[doc_id],
                     filters=filters,
-                    similarity_top_k=10000,
+                    similarity_top_k=Constants.TOP_K,
                 )
             except Exception as e:
                 self.tool.stream_log(
--- a/src/unstract/sdk/index.py
+++ b/src/unstract/sdk/index.py
@@ -13,6 +13,7 @@ from llama_index.core.vector_stores import (
 from typing_extensions import deprecated
 from unstract.adapters.exceptions import AdapterError
 from unstract.adapters.x2text.constants import X2TextConstants
+from unstract.adapters.x2text.dto import TextExtractionResult
 from unstract.adapters.x2text.llm_whisperer.src import LLMWhisperer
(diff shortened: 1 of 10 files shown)


src/unstract/sdk/x2txt.py
```diff
@@ -68,6 +69,16 @@ def process(
         output_file_path: Optional[str] = None,
         **kwargs: dict[Any, Any],
     ) -> str:
+        if kwargs.get("enable_highlight", False) and isinstance(
```
chandrasekharan-zipstack, line 72: @johnyrahul I suggest renaming `enable_highlight` argument instead to something generic like `append_metadata` / `provide_metadata`. Later on for other providers we can try to include support based on what they provide.

src/unstract/sdk/index.py
```diff
@@ -127,6 +127,7 @@ def index(
         reindex: bool = False,
         file_hash: Optional[str] = None,
         output_file_path: Optional[str] = None,
+        enable_highlight: bool = False,
```
chandrasekharan-zipstack, line 130: @johnyrahul I suggest renaming `enable_highlight` argument instead to something generic like `extract_with_metadata`

src/unstract/sdk/x2txt.py
```diff
@@ -68,6 +69,16 @@ def process(
...
+            self._x2text_instance, LLMWhisperer
+        ):
+            output = self._x2text_instance.process_with_hash(
+                input_file_path, output_file_path, **kwargs
+            )
+
+            self._tool.update_exec_metadata({"whisper_hash": output["whisper_hash"]})
```
chandrasekharan-zipstack, line 79: I believe this would create/update a `METADATA.json` in the context of tool execution in a workflow. Do check how this turns out in the context of Prompt Studio

src/unstract/sdk/index.py
johnyrahul: @chandrasekharan-zipstack This argument is not something generic.  This is specifically for the text extraction logic alone.

meta_data will be always there as part of tool run, This flag/argument enables text extraction with whisper has only in case of llm_whisper and  updates whisper hash to metadata,.json

src/unstract/sdk/x2txt.py
johnyrahul: In the context of Prompt Studio, we might have to handle it differently. Text extraction is a sub-step while indexing, so we need to record it somewhere. This might require further discussions.

src/unstract/sdk/index.py
```diff
@@ -246,9 +250,29 @@ def index(
...
+                    if enable_highlight and isinstance(
+                        x2text._x2text_instance, LLMWhisperer
+                    ):
+                        process_response: TextExtractionResult = x2text.process(
+                            input_file_path=file_path,
+                            output_file_path=output_file_path,
+                            enable_highlight=enable_highlight,
+                        )
+                        whisper_hash_value = (
+                            process_response.extraction_metadata.whisper_hash
```
gaya3-zipstack, line 262: For Unstructured Community/Enterprise,  extraction_metadata will be None and accessing whisper_hash from None will raise an exception @johnyrahul

gaya3-zipstack (approved):
Looks good.

muhammad-ali-e (approved):
LGTM

src/unstract/sdk/x2txt.py
johnyrahul: This will require taking it up as a different activity, as the current scope is API deployment alone.

src/unstract/sdk/index.py
```diff
@@ -246,9 +250,29 @@ def index(
...
+                        )
+
+                        metadata = {X2TextConstants.WHISPER_HASH: whisper_hash_value}
+
+                        self.tool.update_exec_metadata(metadata)
```
chandrasekharan-zipstack, line 267: @johnyrahul this would ideally write to `METADATA.json` in the path that's configured for the `Tool` (SDK's base class construct). These were part of the workflow execution related changes and since we're now calling it in the context of `index()` which can be done by `backend` as well - did you check
1.  How we generate / manage this file?
1. Does it overwrite existing contents?
1. I suggest wrapping it within a `extractor_metadata` similar to what we did for `unstract-adapters` to support any such values across extractors generically

Can you update the JIRA with its contents when it runs from prompt studio VS when it runs through a workflow?

johnyrahul: It was working in the prompt studio aspect as well. 
Regarding point 2 It will overwrite the existing keys if passed. Here  it is all together new key `whisper-hash`

johnyrahul: Also, the enable highlight argument is defaulted to a false hence this piece of code won't get executed in the prompt studio aspect.

chandrasekharan-zipstack: Got it - my only other suggestion would be to not keep this key directly as `whisper_hash` in the root of the JSON and instead follow the same structure that we've adopted for the `ExtractionResult` dataclass you had created in adapters.
Later on if we add more fields for LLMwhisperer or support new fields for other extractors we can easily enhance this

chandrasekharan-zipstack (approved):

PR #74, opened 2024-07-18
FEAT: Add Support for Public Indexing and Prompt Run Functionality

tahierhussain:
## What

The changes implemented in the SDK now support SPS for indexing and prompt run functionalities. An optional flag argument called `is_public_call` has been added. This flag primarily determines the following functionalities:

1. **Adapter Configuration**: When a public call is made for indexing or prompt run functionality, the adapter ID will not be available as this is not tenant-specific. If `is_public_call` is `true`, the adapter configuration data will be retrieved from the env variables instead of the database.
2. **Access Restriction**: When a public call is made, tenant details won't be available. If not handled properly, this can result in errors. By using the `is_public_call` flag, the SDK ensures that it does not attempt to access platform keys or other tenant-specific functionalities, preventing such errors.

## Why

The SPS feature requires the `index` and `prompt run` functionalities to be available without authentication, i.e., accessible to public users. To implement this while reusing the same base code, the necessary changes have been made in this PR.

## How

By adding an optional flag argument called `is_public_call` and using it to differentiate between public and private calls.

## Relevant Docs

NA

## Related Issues or PRs

NA

NA

## Notes on Testing

NA

## Screenshots

NA

## Checklist


src/unstract/sdk/index.py
```diff
@@ -388,14 +395,25 @@ def generate_file_id(
...
-            "vector_db_config": ToolAdapter.get_adapter_config(self.tool, vector_db),
-            "embedding_config": ToolAdapter.get_adapter_config(self.tool, embedding),
-            "x2text_config": ToolAdapter.get_adapter_config(self.tool, x2text),
             # Typed and hashed as strings since the final hash is persisted
             # and this is required to be backward compatible
             "chunk_size": str(chunk_size),
             "chunk_overlap": str(chunk_overlap),
         }
+
+        if not self.is_public_call:
```
gaya3-zipstack, line 404: Guess, this will eliminate adapter config combinations from the index_key. Ideally this was used to track any chnage in this combo in order to re-index. Since we are taking this out now, we might be losing out on that. Hope that will not cause a problem @tahierhussain

gaya3-zipstack (approved):
Overall, looks like we are propagating a flag to identify if the call is being made by simple prompt_studio or not.
We could have explored to see if  we could have had a mock org with the adapters configured there and used some kind of an internal platform API service key.

src/unstract/sdk/index.py
gaya3-zipstack: Got it cleared.

chandrasekharan-zipstack, line 404: @tahierhussain what happens to the doc_id when we decide to use different adapters for this simple prompt studio feature? I think the config / JSON itself should be used to obtain the unique ID

chandrasekharan-zipstack (commented):
@tahierhussain rather than introduce elements of SPS (for eg: referring to certain env variables prefixed with `SPS_` or using the variable `is_public_call`) which results in losing the generic nature of our SDK - try to generalize what you're trying to do.

I see that we wish to add support to the SDK classes to not refer the adapter configuration from platform service and instead rely on an env. Since the `adapter_instance_id` is optional in every class, you could alter the default behaviour to hit `platform-service` if the ID is provided, if not check for an env implicitly to obtain the required config. If this env is also not provided, throw an error.
This completely eliminates all mentions SPS in the SDK code

harini-venkataraman (approved):
Like we discussed, let discuss on setting the envs in adapters. 
Otherwise, LGTM.

src/unstract/sdk/index.py
tahierhussain: @chandrasekharan-zipstack As per our earlier discussion, this scenario will now be handled with the latest changes.

src/unstract/sdk/helper.py
```diff
@@ -16,3 +19,24 @@ def get_platform_base_url(platform_host: str, platform_port: str) -> str:
...
+        Returns:
+            bool: True if the adapter_id matches any public adapter key, False otherwise.
+        """
+        try:
+            # Retrieve all attribute names from the PublicAdapterKeys class
+            attributes = dir(PublicAdapterKeys)
+            public_adapter_keys = {getattr(PublicAdapterKeys, attr) for attr in attributes}
+            return adapter_id in public_adapter_keys
+        except Exception:
+            return False
```
chandrasekharan-zipstack, line 42: NIT: You could log any errors here as a warning

src/unstract/sdk/llm.py
```diff
@@ -54,12 +55,17 @@ def _initialise(self):
...
-            platform_api_key = self._tool.get_env_or_die(ToolEnv.PLATFORM_API_KEY)
-            CallbackManager.set_callback(
-                platform_api_key=platform_api_key,
-                model=self._llm_instance,
-                kwargs=self._usage_kwargs,
+
+            is_public_adapter = SdkHelper.is_public_adapter(
+                adapter_id=self._adapter_instance_id
             )
+            if not is_public_adapter:
```
chandrasekharan-zipstack, line 62:
```suggestion
            if not SdkHelper.is_public_adapter(adapter_id=self._adapter_instance_id):
```
Please take care of line length issues if you accept this. You could extend the same to other similar usages as well

chandrasekharan-zipstack (approved):
LGTM - left some minor comments which could be incorporated

chandrasekharan-zipstack:
@tahierhussain please take care of bumping the SDK version as well accordingly

chandrasekharan-zipstack (approved):

src/unstract/sdk/helper.py
chandrasekharan-zipstack: @tahierhussain resolving this - please add it later if you feel its necessary

tahierhussain: @chandrasekharan-zipstack Done. The warning log has been added.

chandrasekharan-zipstack (approved):

PR #75, opened 2024-07-19
[FEATURE] Send provider in the usage callback

Deepak-Kesavan:
## What

Send provider in the usage callback

## Why

It is required for cost calculation

## How

Get the provider from adapter class and set it in `usage_kwargs`

## Relevant Docs

-

## Related Issues or PRs

- https://github.com/Zipstack/unstract-adapters/pull/64

-

## Notes on Testing

...

## Screenshots

...

## Checklist

### DIFF
--- a/src/unstract/sdk/embedding.py
+++ b/src/unstract/sdk/embedding.py
@@ -71,6 +71,7 @@ class Embedding:
             ][Common.ADAPTER]
             embedding_metadata = embedding_config_data.get(Common.ADAPTER_METADATA)
             embedding_adapter_class = embedding_adapter(embedding_metadata)
+            self._usage_kwargs["provider"] = embedding_adapter_class.get_provider()
             return embedding_adapter_class.get_embedding_instance()
         except Exception as e:
             self._tool.stream_log(
--- a/pdm.lock
+++ b/pdm.lock
@@ -5,7 +5,7 @@
 groups = ["default", "docs", "lint", "test"]
 strategy = ["cross_platform"]
 lock_version = "4.4.1"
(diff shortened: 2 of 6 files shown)


gaya3-zipstack (approved):
Looks fine. Aapters 0.22.0 needs to be published for pdm.lock update

src/unstract/sdk/embedding.py
```diff
@@ -71,6 +71,7 @@ def _get_embedding(self) -> BaseEmbedding:
             ][Common.ADAPTER]
             embedding_metadata = embedding_config_data.get(Common.ADAPTER_METADATA)
             embedding_adapter_class = embedding_adapter(embedding_metadata)
+            self._usage_kwargs["provider"] = embedding_adapter_class.get_provider()
```
chandrasekharan-zipstack, line 74: @Deepak-Kesavan I don't remember how we use this usage dictionary - but if we try to generically fetch/ use it for all adapters it'll be good to add the same key for other adapters too. This will allow us to not worry about `KeyError`s while accessing it.
Ignore this comment if we explicitly read this `usage_kwargs` only for embedding and LLM types

chandrasekharan-zipstack (approved):
LGTM

src/unstract/sdk/embedding.py
chandrasekharan-zipstack: I see that we initialize it [here](https://github.com/Zipstack/unstract-sdk/pull/75/files#diff-ae0b5825b9cfb1f4a712504607c8101ae4f9c639a8b1ba08713c5558ba21b36b), ignore this comment

chandrasekharan-zipstack (approved):

PR #88, opened 2024-08-14
Remove tool id from index key id generation

gaya3-zipstack:
## What

Remove tool id from index key id generation

## Why

While generating the doc_id / index key  for indexing, tool id has always been part of it. tool id need not be part of the index key id generation as the indexed documents are not dependent on the tool. Hence it needs to be removed out of the equation

## How

- Deprecate the existing API that takes tool_id as a param while generating index_id
- Introduce a new API without tool id to genertae the index id
- To take care of existing usages of genertae_file_id API, in the implementation, remove the usage of tool_id even though it is passed

## Relevant Docs

-

## Related Issues or PRs

-

-

## Notes on Testing

- For an already indexed doc using the tool_id, after using the new API, it will still go ahead and index it as the index key that generate_index_key is different from the id generated by the now deprecated generate_file_id

## Screenshots

...

## Checklist

### DIFF
--- a/src/unstract/sdk/index.py
+++ b/src/unstract/sdk/index.py
@@ -142,8 +142,7 @@ class Index:
         Returns:
             str: A unique ID for the file and indexing arguments combination
         """
-        doc_id = self.generate_file_id(
-            tool_id=tool_id,
+        doc_id = self.generate_index_key(
             vector_db=vector_db_instance_id,
             embedding=embedding_instance_id,
             x2text=x2text_instance_id,
@@ -335,9 +334,8 @@ class Index:
         finally:
             vector_db.close()
 
-    def generate_file_id(
+    def generate_index_key(
         self,
-        tool_id: str,
         vector_db: str,
         embedding: str,
         x2text: str,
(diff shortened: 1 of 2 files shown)


src/unstract/sdk/index.py
```diff
@@ -388,7 +384,28 @@ def generate_file_id(
         hashed_index_key = ToolUtils.hash_str(json.dumps(index_key, sort_keys=True))
         return hashed_index_key
 
-    @deprecated("Instantiate Index and call index() instead")
+    @deprecated("Instantiate generate_index_key() instead")
```
chandrasekharan-zipstack, line 387: NIT: @gaya3-zipstack we could provide `version` argument, like [so](https://github.com/Zipstack/unstract-sdk/blob/f8a425af2f00be4f6a19a1bf9e117c311d04cfaa/src/unstract/sdk/tool/stream.py#L234) to keep track of when something was deprecated.

chandrasekharan-zipstack (approved):

jaseemjaskp (approved):
LGTM

hari-kuriakose (approved):
@gaya3-zipstack LGTM

src/unstract/sdk/index.py
gaya3-zipstack: That is a great add-on. Done. Thanks @chandrasekharan-zipstack

chandrasekharan-zipstack (approved):

PR #92, opened 2024-08-22
Change token counter implementation & support embedding token counting (defaults to tiktoken)

gaya3-zipstack:
## What

1. Fix token counting for Bedrock to report LLM tokens
2. Support embedding token count recording. However, since mebedding tokens are not provided by the Embedding provider directly unlike LLM tokens, we are using llama-index's default token counting using tiktoken

## Why

1. Reporting LLM tokens for Bedrock
2. Embedding token counting was missing in some flows like indexing and RAG use cases

## How

1. Upgrade Bedrock provider lib to the latest with the fix for reporting token count usage
2. Wire in the callback_manager to the llama-index calls doing the indexing

## Relevant Docs

-

## Related Issues or PRs

-

-

## Notes on Testing

Attaching the token usage table csv
[token_usage.csv](https://github.com/user-attachments/files/16715432/token_usage.csv)
 file generated out of testing supported LLMs and Embeddings

## Screenshots

...

## Checklist

### DIFF
--- a/src/unstract/sdk/utils/token_counter.py
+++ b/src/unstract/sdk/utils/token_counter.py
@@ -27,19 +27,17 @@ class TokenCounter:
         completion_tokens = Constants.DEFAULT_TOKEN_COUNT
         if EventPayload.PROMPT in payload:
             response = payload.get(EventPayload.COMPLETION)
-            token_counts: dict[str, int] = TokenCounter.get_tokens_from_response(
-                response
-            )
-            prompt_tokens = int(token_counts["prompt_tokens"])
-            completion_tokens = int(token_counts["completion_tokens"])
+            (
+                prompt_tokens,
+                completion_tokens,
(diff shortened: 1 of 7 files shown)


src/unstract/sdk/utils/token_counter.py
gaya3-zipstack, line 1: @hari-kuriakose I am removing a private function in this file and introducing a new one. Since it is a private function, I hope I need not work on backward compatibility piece of the removed function.

chandrasekharan-zipstack (approved):

kirtimanmishrazipstack (approved):
LGTM

hari-kuriakose (commented):
@gaya3-zipstack LGTM overall.

Since we are refactoring the token counting logic to be cleaner and have more coverage, we can definitely proceed.

However in future iterations, let's do add unit tests here, for obvious reasons.

PR #95, opened 2024-08-30
[FEATURE] Added function to get model name of LLM

Deepak-Kesavan:
## What

Added function to get model name of LLM

## Why

To make it easier to get the model name

## How

...

## Relevant Docs

-

## Related Issues or PRs

-

-

## Notes on Testing

...

## Screenshots

...

## Checklist

### DIFF
--- a/src/unstract/sdk/__init__.py
+++ b/src/unstract/sdk/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "0.46.0"
+__version__ = "0.46.1"
 
 
 def get_sdk_version():
--- a/src/unstract/sdk/__init__.py
+++ b/src/unstract/sdk/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "0.46.1"
+__version__ = "0.47.0"
 
 
 def get_sdk_version():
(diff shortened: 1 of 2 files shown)


src/unstract/sdk/__init__.py
```diff
@@ -1,4 +1,4 @@
-__version__ = "0.46.0"
+__version__ = "0.46.1"
```
Deepak-Kesavan, line 1: Bumped only patch version because it is a small change

chandrasekharan-zipstack: Since its a new function - ideally it should be a minor version bump. Approving to unblock anyway

chandrasekharan-zipstack (approved):

src/unstract/sdk/__init__.py
Deepak-Kesavan: I also thought the same. Since I planned to update only prompt service, made it as a patch.

Let me update minor version itself.

gaya3-zipstack (approved):
Looks good

PR #96, opened 2024-08-30
fix: Introduced new index_document() to fix chunking related issue

chandrasekharan-zipstack:
## What

- Deprecated old function `get_vector_store_index_from_storage_context()` to index and created a new one `index_document()` in `vector_db.py`
- Moved attributes within new function to simplify logic at caller side
- Renamed deprecated `SimpleNodeParser` to the intended `SentenceSplitter`
- Bumped SDK to `0.48.0`

## Why

- The llama index function `VectorStoreIndex.from_documents()` expects an attribute `transformations` which was not set correctly. As a result the user configured values for chunk size and chunk overlap was never set and used

## Notes on Testing

- Attempted indexing locally

## Screenshots

1. With 500 and 50 chunk size / overlap
image
1. With 2048 and 256 chunk size / overlap
image

## Checklist

### DIFF
--- a/src/unstract/sdk/__init__.py
+++ b/src/unstract/sdk/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "0.47.0"
+__version__ = "0.48.0"
 
 
 def get_sdk_version():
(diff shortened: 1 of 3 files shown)


harini-venkataraman (approved):

Deepak-Kesavan (approved):

gaya3-zipstack (approved):
Looks good

PR #98, opened 2024-09-02
[FIX] JSON regex check based on param

Deepak-Kesavan:
## What

JSON regex check based on param

## Why

The JSON regex check is based on matching characters like `[,]` or `{,}`. If there are pairs of these brackets within the response, the function will extract only the content between the matching brackets even if it is not a JSON.

## How

Added a parameter `extract_json`, which uses a regex to extract content only if it is set to True

## Relevant Docs

-

## Related Issues or PRs

-

-

## Notes on Testing

...

## Screenshots

...

## Checklist

### DIFF
--- a/src/unstract/sdk/llm.py
+++ b/src/unstract/sdk/llm.py
@@ -69,7 +69,7 @@ class LLM:
     def complete(
         self,
         prompt: str,
-        extract_json: bool = False,
+        extract_json: bool = True,
         process_text: Optional[Callable[[str], str]] = None,
         **kwargs: Any,
     ) -> dict[str, Any]:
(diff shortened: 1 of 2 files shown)


src/unstract/sdk/llm.py
```diff
@@ -69,24 +69,28 @@ def _initialise(self):
     def complete(
         self,
         prompt: str,
+        extract_json: bool = True,
```
Deepak-Kesavan, line 72: Kept default as true to retain the previous behavior.

chandrasekharan-zipstack (approved):

harini-venkataraman (approved):

PR #100, opened 2024-09-12
[FIX] Delete Vector DB nodes after extraction

Deepak-Kesavan:
## What

- Delete Vector DB nodes after extraction

## Why

- If same file in workflow, reindex of one call deletes the vector db node if found and then proceeds extraction. Since extraction is a time consuming process existing node should not be deleted before extraction. Else during this time, if we try to read the vector db the node won't be available.

## How

- Delete the node only after extraction in case of `reindex=True`.

## Relevant Docs

-

## Related Issues or PRs

-

-

## Notes on Testing

...

## Screenshots

...

## Checklist

### DIFF
--- a/src/unstract/sdk/index.py
+++ b/src/unstract/sdk/index.py
@@ -203,20 +203,7 @@ class Index:
                     level=LogLevel.ERROR,
                 )
 
-            if doc_id_found and reindex:
-                # Delete the nodes for the doc_id
-                try:
-                    vector_db.delete(ref_doc_id=doc_id)
-                    self.tool.stream_log(f"Deleted nodes for {doc_id}")
-                except Exception as e:
-                    self.tool.stream_log(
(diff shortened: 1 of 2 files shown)


src/unstract/sdk/index.py
```diff
@@ -287,6 +274,18 @@ def index(
                 documents.append(document)
             self.tool.stream_log(f"Number of documents: {len(documents)}")
 
+            if doc_id_found:
```
ritwik-g, line 277: @Deepak-Kesavan if the doc_id_found is what happens? We will still index right? May be add a comment here to make it clear.

ritwik-g (approved):

chandrasekharan-zipstack (approved):

src/unstract/sdk/index.py
Deepak-Kesavan: if doc_id_found we delete and index, else we just index alone.

PR #106, opened 2024-10-01
feat: Changes to measure time taken in index

chandrasekharan-zipstack:
## What

- Wrapper to log time elapsed in a function call
- Refactored indexing call to measure extraction separately

## Why

- To gain some insights on long running calls

## Notes on Testing

- Tried running some prompts locally and tried running a workflow

## Screenshots
- Logs from a workflow run
image

## Checklist

I have read and understood the [Contribution Guidelines]().

PR #107, opened 2024-10-01
feat: Measure indexing time separately

chandrasekharan-zipstack:
## What

- Refactored to measure time to index separately
- Measuring time for answer prompt API

## Why

- Gather insight on time spent on some calls

## Related Issues or PRs

- https://github.com/Zipstack/unstract-sdk/pull/106

## Notes on Testing

- Tried indexing a doc in prompt studio

## Screenshots

image

## Checklist

### DIFF
--- a/src/unstract/sdk/index.py
+++ b/src/unstract/sdk/index.py
@@ -387,7 +387,7 @@ class Index:
             raise IndexingError(str(e)) from e
 
         self.tool.stream_log("File has been indexed successfully")
-        return doc_id
+        return
 
     def generate_index_key(
         self,
--- a/src/unstract/sdk/index.py
+++ b/src/unstract/sdk/index.py
@@ -178,7 +178,7 @@ class Index:
                 logger.error(f"Error occured inside function 'process_text': {e}")
         return extracted_text
 
-    @log_elapsed(operation="INDEXING(might include EXTRACTION)")
+    @log_elapsed(operation="CHECK_AND_INDEX(overall)")
     def index(
         self,
         tool_id: str,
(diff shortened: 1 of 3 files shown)


Deepak-Kesavan (approved):

src/unstract/sdk/index.py
```diff
@@ -293,82 +293,101 @@ def index(
...
+                embedding=embedding,
+                chunk_size=chunk_size,
+                chunk_overlap=chunk_overlap,
+                doc_id=doc_id,
+                text_to_idx=extracted_text,
+                doc_id_found=doc_id_found,
+            )
+            return doc_id
+        finally:
+            vector_db.close()
```
ritwik-g, line 307: @chandrasekharan-zipstack if we return in line 305 we will never even reach this line. How will the vector DB be closed in such case? Shouldn't this be included inisde try as well?

Also there isn't even a catch block. How will this behave in case of error?

ritwik-g (commented):
@chandrasekharan-zipstack I have one comment https://github.com/Zipstack/unstract-sdk/pull/107/files#r1782255338

But I don't have enough context on the refactoring and logic happening so please get the approval from folks who have more knowledge on this area. May be @gaya3-zipstack or @harini-venkataraman can help

src/unstract/sdk/index.py
chandrasekharan-zipstack: @ritwik-g the `finally` block purely takes care of closing the vector DB connection in case of any error. The reason its added here is because [this](https://github.com/Zipstack/unstract-sdk/blob/4ee2c3f46c453fdaf404d2c899e01dc15486a3e3/src/unstract/sdk/index.py#L244-L248) is where we create the connection as well. In spite of the return, I believe the logic in the `finally` block gets executed

src/unstract/sdk/index.py
```diff
@@ -293,82 +293,101 @@ def index(
...
+    def index_to_vector_db(
+        self,
+        vector_db: VectorDB,
+        embedding: Embedding,
+        chunk_size: int,
+        chunk_overlap: int,
+        text_to_idx: str,
+        doc_id: str,
+        doc_id_found: bool,
+    ):
```
gaya3-zipstack, line 319: index_to_vector_db returns doc_id (str). EIther we can not return it or change signature to return str. Isn't mypy capturing these ?

chandrasekharan-zipstack: Addressed this

gaya3-zipstack (approved):
Looks good.

src/unstract/sdk/index.py
gaya3-zipstack: Yes finally block executes the last both in case of exceptions and normal return flows.

PR #112, opened 2024-10-17
File Storage interface and implementation

gaya3-zipstack:
## What

File Storage interface and implementation using fsspec

## Why

To support remote storage for k8s

## How

A simple file operationinterface has been designed and the same has been implemented using fsspec

## Relevant Docs

- 

## Related Issues or PRs

- Platform Service - https://github.com/Zipstack/unstract/pull/811

-

## Notes on Testing

Unit test cases have been added under tests/test_file_storage.py. For env variables, pls refer to sample.env under tests. The newly added variables can be copied and updated accordingly for running the tests.

## Screenshots

...

## Checklist


src/unstract/sdk/file_storage/fs_mode.py
hari-kuriakose, line 1: @gaya3-zipstack Please remove this from here. File storage mode is not the responsibility of this layer at all.

Access modes need to be taken care of downstream.

src/unstract/sdk/file_storage/fs_type.py
```diff
@@ -0,0 +1,10 @@
+import enum
+
+
+class FileStorageType(enum.Enum):
```
hari-kuriakose, line 4: @gaya3-zipstack More accurate name will be `FileStorageProvider`.

src/unstract/sdk/file_storage/fs_interface.py
```diff
@@ -0,0 +1,61 @@
...
+    @abstractmethod
+    def open(
+        self,
+        path: str,
+        mode: str,
+        encoding: str = Common.DEFAULT_ENCODING,
+    ) -> Union[AbstractFileSystem]:
+        pass
+
+    def read(
```
hari-kuriakose, line 19: @gaya3-zipstack Won't we need `abstractmethod` decorator here too?

src/unstract/sdk/file_storage/fs_impl.py
```diff
@@ -0,0 +1,180 @@
...
+from unstract.sdk.file_storage.fs_interface import FileStorageInterface
+from unstract.sdk.file_storage.helper import FileStorageHelper
+
+
+class FileStorage(FileStorageInterface):
+    # This class integrates fsspec library for file operations
+
+    fs: fsspec  # fsspec file system handle
+
+    def __init__(self, type, credentials: dict() = {}):
```
hari-kuriakose, line 17: @gaya3-zipstack More accurate variable name will be `provider` instead of `type`.
All are different provider for same file type storage.

src/unstract/sdk/file_storage/helper.py
```diff
@@ -0,0 +1,55 @@
...
+        try:
+            os.environ[GCS.CREDS] = credentials[GCS.CREDS]
+            storage_client = storage.Client()
+
+            fs = fsspec.filesystem(
+                protocol=FileStorageType.GCS.value,
+                project=credentials[GCS.PROJECT_ID],
+                client=storage_client,
+            )
+            logger.info(f"Connected to {FileStorageType.GCS.value} file system")
```
hari-kuriakose, line 33: @gaya3-zipstack Please change to debug log level everywhere.

src/unstract/sdk/file_storage/fs_impl.py
gaya3-zipstack: Done.

src/unstract/sdk/file_storage/fs_interface.py
gaya3-zipstack: Yes I missed this. Added now. Thanks!

src/unstract/sdk/file_storage/fs_mode.py
gaya3-zipstack: Got it. Removed.

src/unstract/sdk/file_storage/fs_interface.py
ritwik-g: @gaya3-zipstack is the latest code not pushed. Don't see the decorator here.

src/unstract/sdk/file_storage/helper.py
```diff
@@ -0,0 +1,55 @@
...
+    def file_storage_init(type: FileStorageType, credentials):
+        if type == FileStorageType.GCS:
+            return FileStorageHelper.gcs_init(credentials)
+        elif type == FileStorageType.Local:
+            return FileStorageHelper.local_file_system_init()
+
+    @staticmethod
+    def gcs_init(credentials):
+        try:
+            os.environ[GCS.CREDS] = credentials[GCS.CREDS]
```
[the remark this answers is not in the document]
ritwik-g: @gaya3-zipstack also am not sure why we need to set this in `os.environ`? I mean aren't we already expecting to get this from the env?

src/unstract/sdk/file_storage/helper.py
```diff
@@ -0,0 +1,55 @@
...
+            storage_client = storage.Client()
+
+            fs = fsspec.filesystem(
+                protocol=FileStorageType.GCS.value,
+                project=credentials[GCS.PROJECT_ID],
+                client=storage_client,
+            )
+            logger.info(f"Connected to {FileStorageType.GCS.value} file system")
+        except Exception as e:
+            logger.error(
```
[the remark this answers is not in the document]
ritwik-g: @gaya3-zipstack also please add the exc_info and stack_info in logger.error so that the entire error trace will be printed. I think that would be needed for debugging.

src/unstract/sdk/file_storage/helper.py
```diff
@@ -0,0 +1,55 @@
...
+                f"Error in initialising {FileStorageType.GCS.value} file system {e}"
+            )
+            raise FileStorageError(str(e))
+        return fs
+
+    @staticmethod
+    def azure_init(extra_config):
+        return
```
ritwik-g, line 43: @gaya3-zipstack instead of return I think it would be better raise an error with message `NotImplemented` till we implement this

src/unstract/sdk/file_storage/fs_impl.py
ritwik-g, line 17: @gaya3-zipstack where will these credentials and type be passed from in our final implementation.

src/unstract/sdk/file_storage/helper.py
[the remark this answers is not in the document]
ritwik-g: Also couldn't understand the reason why we need to assign this back to ENV

src/unstract/sdk/file_storage/fs_impl.py
```diff
@@ -0,0 +1,180 @@
...
+            Union[AbstractFileSystem] : Actual file handle to the file that was opened
+        """
+        try:
+            return self.fs.open(path=path, mode=mode, encoding=encoding)
+        except Exception as e:
+            raise FileOperationError(str(e))
+
+    def read(
+        self,
+        file_handle: Union[AbstractFileSystem],
```
ritwik-g, line 47: @hari-kuriakose @gaya3-zipstack the methods where we are using file_handle. Am just thinking is there any real benefit to making this an input? Couldn't we make this an abstraction and make the file_handle an instance variable?

tests/fsspec-test/input/2.pdf
ritwik-g, line 1: @gaya3-zipstack any specific reason why we need pdf for this test case? Wouldn't txt be enough or is there any special case?

Usually adding binaries inside a git repo will make the overall repo size larger. So it is something we need to avoid when possible. Adding a file which won't change much is fine. Pointing it out incase you are not aware of this

ritwik-g (commented):
@gaya3-zipstack overall looks good. But check the comments from @hari-kuriakose mainly. Mine are more open ended so we can discuss during the call.

Other than that please add doc strings to all the new class methods and classes. Since this is a new one let's add and maintain from the beginning.

src/unstract/sdk/file_storage/fs_interface.py
gaya3-zipstack: Yea, I was just addressing and replying here for tracking. Pushed latest a bit back.

src/unstract/sdk/file_storage/helper.py
gaya3-zipstack: Done.

src/unstract/sdk/file_storage/helper.py
gaya3-zipstack: Done

src/unstract/sdk/file_storage/helper.py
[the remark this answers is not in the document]
gaya3-zipstack: @hari-kuriakose @ritwik-g @chandrasekharan-zipstack 
Currently the env  variable for FileStorage in encapsulated iand looks like FILE_STORAGE='{"GOOGLE_APPLICATION_CREDENTIALS": "/home/gayathrivijayakumar/code/Unstract/unstract-sdk/tests/google-creds.json","PROJECT_ID": "pandoras-tamer"}'
The contents of FILE_STORAGE will vary depending on the storage provider.  

I was initialising GCSFS by setting the env GOOGLE_APPLICATION_CREDENTIALS. However, just learnt that we could do it other ways also. Thank you for leading me to this.
Anyway I have now changed it to pass the creds through function args. This concern should hence be addressed. Pls let me know if something looks odd.

src/unstract/sdk/file_storage/fs_impl.py
```diff
@@ -0,0 +1,178 @@
...
+from unstract.sdk.file_storage.fs_interface import FileStorageInterface
+from unstract.sdk.file_storage.helper import FileStorageHelper
+
+
+class FileStorage(FileStorageInterface):
+    # This class integrates fsspec library for file operations
+
+    fs: fsspec  # fsspec file system handle
+
+    def __init__(self, provider, credentials: dict() = {}):
```
ritwik-g, line 17: @gaya3-zipstack do not initiate the dict like this. This will lead to unexpected issues. Keep it None as default.

src/unstract/sdk/file_storage/fs_permanent.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+class PermanentFileStorage(FileStorage):
+    SUPPORTED_FILE_STORAGE_TYPES = [
+        FileStorageProvider.GCS.value,
+        FileStorageProvider.S3.value,
+        FileStorageProvider.Azure.value,
+    ]
+
+    def __init__(self, provider, credentials: dict() = None):
```
ritwik-g, line 18: @gaya3-zipstack is adding type hint with brackets correct? `dict()` . I haven't observed this before hence asking

src/unstract/sdk/file_storage/fs_permanent.py
```diff
@@ -0,0 +1,42 @@
...
+        if provider.value not in self.SUPPORTED_FILE_STORAGE_TYPES:
+            raise FileStorageError(
+                f"File storage provider is not supported in Permanent mode. "
+                f"Supported providers: {self.SUPPORTED_FILE_STORAGE_TYPES}"
+            )
+        if provider == FileStorageProvider.GCS:
+            super().__init__(provider, credentials)
+        else:
+            raise FileStorageError("File system not supported")
```
ritwik-g, line 27: How is the first if and second if block different. Could you may be add some comments?

src/unstract/sdk/file_storage/helper.py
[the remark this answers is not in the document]
chandrasekharan-zipstack: @hari-kuriakose @gaya3-zipstack @ritwik-g do we wish to log the error with the traceback in the SDK? If we're treating it as a library, I suggest we just raise the exception alone and let the downstream handle the logging. This can help avoid repeated tracebacks
- We've already added some logging middlewares in most services which log the traceback itself. 
- This might be missing in our celery workers alone which can be supported as well with a try-except on the shared task

src/unstract/sdk/file_storage/fs_permanent.py
chandrasekharan-zipstack: Please run the mypy checks once @gaya3-zipstack - I think you might have to change the type annotation as `dict[str, Any]`. Also please annotate the `provider` and the return type as well

src/unstract/sdk/file_storage/helper.py
```diff
@@ -0,0 +1,84 @@
...
+
+        Args:
+            credentials (dict): credentials
+
+        Returns:
+            NA
+        """
+        try:
+            gcs_creds = service_account.Credentials.from_service_account_file(
+                credentials[GCS.CREDS],
```
chandrasekharan-zipstack, line 42: @gaya3-zipstack this can lead to `KeyError`s if this env is not properly intitialized - where do we validate this `credentials` dict?
NIT: Please add type hints to all methods

tests/test_file_storage.py
```diff
@@ -0,0 +1,407 @@
...
+    file_storage = FileStorage(provider=provider, credentials=credentials)
+    assert file_storage is not None
+    return file_storage
+
+
+def permanent_file_storage(provider: FileStorageProvider):
+    credentials = json.loads(os.environ.get(TEST_CONSTANTS.FILE_STORAGE_ENV))
+    file_storage = PermanentFileStorage(provider=provider, credentials=credentials)
+    assert file_storage is not None
+    return file_storage
```
chandrasekharan-zipstack, line 39: NIT: You can add these with a `pytest.fixture` decorator in `conftest.py` to reuse them in your tests rather than explicitly creating them in `paramterize()`. You can refer [these tests](https://github.com/Zipstack/llm-whisperer-python-client/blob/c97cb4582d9dade85788e894f4060d5cec100b31/tests/conftest.py#L10) for reference

tests/test_file_storage.py
```diff
@@ -0,0 +1,407 @@
...
+    read_length,
+    expected_write_length,
+):
+    if file_storage.exists(path=TEST_CONSTANTS.WRITE_FOLDER_PATH):
+        file_storage.rm(path=TEST_CONSTANTS.WRITE_FOLDER_PATH, recursive=True)
+    assert file_storage.exists(path=TEST_CONSTANTS.WRITE_FOLDER_PATH) == False  # noqa
+    file_storage.mkdir(path=TEST_CONSTANTS.WRITE_FOLDER_PATH)
+    # assert file_storage.exists(path=TEST_CONSTANTS.WRITE_FOLDER_PATH) == True
+    if read_file_path:
+        file_read_handle = file_storage.open(path=read_file_path, mode=read_mode)
```
chandrasekharan-zipstack, line 197: NIT: The file handle might not be closed here since we need to explicitly close it

src/unstract/sdk/file_storage/fs_impl.py
```diff
@@ -0,0 +1,178 @@
...
+        self.fs = FileStorageHelper.file_storage_init(
+            provider=provider, credentials=credentials
+        )
+
+    def open(
```
chandrasekharan-zipstack, line 22: @gaya3-zipstack I suggest making use of context managers instead to cleanly open and close files - it'll be a pain to keep track / explicitly close the file handles by the caller

src/unstract/sdk/file_storage/helper.py
gaya3-zipstack: Yes @chandrasekharan-zipstack  There will be a KeyError that will be caught by the excpetion and printed.

src/unstract/sdk/file_storage/fs_permanent.py
gaya3-zipstack: Second if is actually to catch non-implemented types. Changed the exception now to keep that clear..

src/unstract/sdk/file_storage/fs_permanent.py
gaya3-zipstack: Addressed. Thank you. I forgot that mypy is disabled.

src/unstract/sdk/file_storage/fs_impl.py
ritwik-g: This works by reference and this variable will be only created once and basically it will be shared between different calls to this method

ritwik-g: Adding a sample code below to make this clear. Added list also for example. This is applicable for all data types like, dict, list , set etc .

```python
import random
import string

def test_dict(input_dict={}, input_list=[]):
    key = "".join(random.choices(string.ascii_letters, k=7))
    input_dict[key] = True
    input_list.append(key)
    return input_dict, input_list

print(test_dict())
print(test_dict())
print(test_dict())
```

src/unstract/sdk/file_storage/fs_impl.py
```diff
@@ -0,0 +1,157 @@
...
+    # This class integrates fsspec library for file operations
+
+    fs: fsspec  # fsspec file system handle
+
+    def __init__(self, provider, credentials: dict[str, Any] = {}):
+        self.fs = FileStorageHelper.file_storage_init(
+            provider=provider, credentials=credentials
+        )
+
+    def read_file(
```
hari-kuriakose, line 21: @gaya3-zipstack We call this `read` itself.

src/unstract/sdk/file_storage/fs_impl.py
```diff
@@ -0,0 +1,157 @@
...
+        self,
+        path: str,
```
hari-kuriakose, line 23: @gaya3-zipstack What happens if a `dir` path is passed here?

src/unstract/sdk/file_storage/fs_impl.py
```diff
@@ -0,0 +1,157 @@
...
+            encoding (str): Encoding type like utf-8 or utf-16
+            seek_position (int): Position to start reading from
+            length (int): Number of bytes to be read. Default is full
+            file content.
+
+        Returns:
+            Union[bytes, str] - File contents in bytes/string based on the opened mode
+        """
+        try:
+            with self.fs.open(path=path, mode=mode, encoding=encoding) as file_handle:
```
hari-kuriakose, line 44: @gaya3-zipstack What if a very large file is being read? Can we stream the contents back instead of keeping all of it in memory?

src/unstract/sdk/file_storage/fs_impl.py
```diff
@@ -0,0 +1,157 @@
...
+                if seek_position > 0:
+                    file_handle.seek(seek_position)
+                return file_handle.read(length)
+        except Exception as e:
+            raise FileOperationError(str(e))
+
+    def write_file(
```
hari-kuriakose, line 51: @gaya3-zipstack We can name this `write` itself.

src/unstract/sdk/file_storage/fs_impl.py
```diff
@@ -0,0 +1,157 @@
...
+                        include r, rb, w and wb
+            encoding (str): Encoding type like utf-8 or utf-16
+            seek_position (int): Position to start writing from
+            data (Union[bytes, str]): Contents to be written
+
+        Returns:
+            int: Number of bytes that were successfully written to the file
+        """
+        try:
+            with self.fs.open(path=path, mode=mode, encoding=encoding) as file_handle:
```
hari-kuriakose, line 73: @gaya3-zipstack Similarly, what if a very large file is written to? Can we support streaming writes?

src/unstract/sdk/file_storage/fs_impl.py
```diff
@@ -0,0 +1,157 @@
...
+                return file_handle.write(data)
+        except Exception as e:
+            raise FileOperationError(str(e))
+
+    def seek_file(
```
hari-kuriakose, line 78: @gaya3-zipstack We can call this `seek` itself.

src/unstract/sdk/file_storage/fs_impl.py
```diff
@@ -0,0 +1,157 @@
...
+        Returns:
+            int: file pointer location after seeking to the mentioned position
+        """
+        try:
+            with self.fs.open(path=path, mode="rb") as file_handle:
+                return file_handle.seek(location, position)
+        except Exception as e:
+            raise FileOperationError(str(e))
+
+    def make_dir(self, path: str, create_parents: bool = True):
```
hari-kuriakose, line 103: @gaya3-zipstack We can call this `mkdir` itself, as per age old tradition.

src/unstract/sdk/file_storage/fs_impl.py
```diff
@@ -0,0 +1,157 @@
...
+            path (str): Path of the directory to be created
+            create_parents (bool): Specify if parent directories to be created
+            if any of the nested directory does not exist
+        """
+        try:
+            self.fs.mkdir(path, create_parents=create_parents)
+        except Exception as e:
+            raise FileOperationError(str(e))
+
+    def path_exists(self, path: str) -> bool:
```
hari-kuriakose, line 116: @gaya3-zipstack We can call this `exists` itself.

src/unstract/sdk/file_storage/fs_impl.py
```diff
@@ -0,0 +1,157 @@
...
+        """List the directory path.
+
+        Args:
+            path (str): Directory path
+
+        Returns:
+            List[str]: List of files / directories under the path
+        """
+        try:
+            return self.fs.ls(path)
```
hari-kuriakose, line 140: @gaya3-zipstack What happens if a file path is passed here?

src/unstract/sdk/file_storage/fs_impl.py
```diff
@@ -0,0 +1,157 @@
...
+
+        Args:
+            path (str): Path to the file / directory
+            recursive (bool): Whether the files and folders nested
+            under path are to be removed or not
+
+        Returns:
+        """
+        try:
+            return self.fs.rm(path, recursive=recursive)
```
hari-kuriakose, line 155: @gaya3-zipstack
1. What happens if a dir path is passed and recursive is False?
2. Also, what happens if a very large directory is triggered for delete?

src/unstract/sdk/file_storage/fs_impl.py
hari-kuriakose, line 1: @gaya3-zipstack Instead of catching the blanket `Exception`, we need to ideally catch specific exceptions.

src/unstract/sdk/file_storage/fs_permanent.py
```diff
@@ -0,0 +1,63 @@
...
+            file content.
+
+        Returns:
+            Union[bytes, str] - File contents in bytes/string based on the opened mode
+        """
+        try:
+            if not self.path_exists(path):
+                local_file_storage = FileStorage(provider=FileStorageProvider.Local)
+                if local_file_storage.path_exists(path):
+                    self.copy_on_write(path, path)
```
hari-kuriakose, line 60: @gaya3-zipstack All of lines 57-60 is the "copy on write" strategy. Probably we can add a dev comment regarding the same.

This function can be named just `copy` itself.

src/unstract/sdk/file_storage/fs_impl.py
```diff
@@ -0,0 +1,157 @@
...
+                if seek_position > 0:
+                    file_handle.seek(seek_position)
+                return file_handle.read(length)
+        except Exception as e:
+            raise FileOperationError(str(e))
+
+    def write_file(
+        self,
+        path: str,
+        mode: str,
```
hari-kuriakose, line 54: @gaya3-zipstack What happens if we are passing read intended modes here?

src/unstract/sdk/file_storage/fs_permanent.py
```diff
@@ -0,0 +1,63 @@
...
+                f"File storage provider is not supported in Permanent mode. "
+                f"Supported providers: {self.SUPPORTED_FILE_STORAGE_TYPES}"
+            )
+        if provider == FileStorageProvider.GCS:
+            super().__init__(provider, credentials)
+        else:
+            raise NotImplementedError
+
+    def copy_on_write(self, path1, path2):
+        self.fs.put(path1, path2)
```
hari-kuriakose, line 32: @gaya3-zipstack We could expose a `copy` API also in the interface, as that could be generic and stand alone too.

The lines 57-60 below collectively is actually the "copy on write" strategy.

src/unstract/sdk/file_storage/fs_impl.py
ritwik-g: @hari-kuriakose I don't think opening a file in itself will load everything in to the memory. But that being said down below we might be able to improve the read operation.

src/unstract/sdk/file_storage/fs_impl.py
```diff
@@ -0,0 +1,157 @@
...
+            file content.
+
+        Returns:
+            Union[bytes, str] - File contents in bytes/string based on the opened mode
+        """
+        try:
+            with self.fs.open(path=path, mode=mode, encoding=encoding) as file_handle:
+                if seek_position > 0:
+                    file_handle.seek(seek_position)
+                return file_handle.read(length)
```
ritwik-g, line 47: @hari-kuriakose I think the read will load the complete file in to the memory. One way to avoid this is using read lines and yielding each line. But again at the downstream where ever it is being used if the data is not handled in proper manner entire read data could be loaded in to the memory. And in our case I don't know if implementing such a thing will help because I anticipate that all the downstream usage will be needing the entire data in memory.

src/unstract/sdk/file_storage/fs_impl.py
ritwik-g: @hari-kuriakose I would suggest building a separate method for such use cases. Same with read also if needed.

src/unstract/sdk/file_storage/fs_impl.py
ritwik-g: @hari-kuriakose normal `ls` command will display the single files path in linux in this case. Similar is the behaviour with fsspec also.

image

tests/test_file_storage.py
gaya3-zipstack: I tried that. However my case is slightly different where I need to call the fixture itself with different values. When I tried doing that by keeping the file_storage and permanent_file_storage as fixtures, I see an error. I am not investing much time here for now as this is  improvisation. Will come back to it when I get a chance.

src/unstract/sdk/file_storage/fs_impl.py
gaya3-zipstack: Done.

src/unstract/sdk/file_storage/fs_impl.py
gaya3-zipstack: DOne

src/unstract/sdk/file_storage/fs_impl.py
gaya3-zipstack: The behaviour will be controlled by the underlying fsspec implementation.

src/unstract/sdk/file_storage/fs_impl.py
gaya3-zipstack: Done

src/unstract/sdk/file_storage/fs_impl.py
gaya3-zipstack: Done

src/unstract/sdk/file_storage/fs_impl.py
gaya3-zipstack: Added a test case and works as mentioned by @ritwik-g

src/unstract/sdk/file_storage/fs_impl.py
gaya3-zipstack: For (1), we get an error saying recusrive needs to be set to true if the folder to be deleted has nested files/folders under it.

src/unstract/sdk/file_storage/fs_permanent.py
gaya3-zipstack: Done.

src/unstract/sdk/file_storage/helper.py
```diff
@@ -0,0 +1,85 @@
...
+        Returns:
+            NA
+        """
+        if provider == FileStorageProvider.GCS:
+            return FileStorageHelper.gcs_init(credentials)
+        elif provider == FileStorageProvider.Local:
+            return FileStorageHelper.local_file_system_init()
+
+    @staticmethod
+    def gcs_init(credentials):
```
harini-venkataraman, line 32: @gaya3-zipstack Looks like the pre-commit is not run over this commit. Please add the return type to method

src/unstract/sdk/index.py
```diff
@@ -416,6 +422,7 @@ def generate_index_key(
         chunk_overlap: str,
         file_path: Optional[str] = None,
         file_hash: Optional[str] = None,
+        fs: FileStorage = FileStorage(provider=FileStorageProvider.Local),
```
harini-venkataraman, line 425: @gaya3-zipstack QQ on the usage - We have initialed to Local storage here, how do we distinguish between the providers?

src/unstract/sdk/file_storage/fs_impl.py
```diff
@@ -0,0 +1,187 @@
...
+        Returns:
+            int: file pointer location after seeking to the mentioned position
+        """
+        try:
+            with self.fs.open(path=path, mode="rb") as file_handle:
+                return file_handle.seek(location, position)
+        except Exception as e:
+            raise FileOperationError(str(e))
+
+    def mkdir(self, path: str, create_parents: bool = True):
```
harini-venkataraman, line 103: @gaya3-zipstack Can we add `is_exisit` param in the base method?

tests/sample.env
```diff
@@ -10,3 +10,15 @@ EMBEDDING_TEST_VALUES=["", "", ""]
...
+WRITE_FOLDER_PATH=fsspec-test/output
+RECURSION_FOLDER_PATH=fsspec-test/output/recursion-test
+READ_PDF_FILE=fsspec-test/input/2.pdf
+READ_TEXT_FILE=fsspec-test/input/1.txt
+WRITE_PDF_FILE=fsspec-test/output/2.pdf
+WRITE_TEXT_FILE=fsspec-test/output/1.txt
+TEST_FOLDER=fsspec-test/test-folder
+GCS_BUCKET=fsspec-test/input
+TEXT_CONTENT=Writing directly from string as read_file is not passed
+FILE_STORAGE='{"GOOGLE_APPLICATION_CREDENTIALS": "/../google-creds.json","PROJECT_ID": "abcdef"}'
```
harini-venkataraman, line 24: @gaya3-zipstack Like discussed, changing the env to `FILE_STORAGE_PROVIDER` & `FILE_STORAGE_CREDENTIAL` in OSS, so that we can identify the provider used. Adding it here for tracking.
CC : @hari-kuriakose

src/unstract/sdk/utils/tool_utils.py
```diff
@@ -50,13 +55,17 @@ def get_hash_from_file(file_path: str) -> str:
         h = sha256()
         b = bytearray(128 * 1024)
         mv = memoryview(b)
-        with open(file_path, "rb", buffering=0) as f:
-            while n := f.readinto(mv):
-                h.update(mv[:n])
+
+        file_contents = fs.read(path=file_path, mode="rb")
+        mv = memoryview(file_contents)
+        h.update(mv[: len(file_contents)])
```
hari-kuriakose, line 61: @gaya3-zipstack This will work, but not efficient. This is one of the downsides of not exposing the file handle explicitly.

1. `buffering=0` is required to disable buffering.
2. `memoryview` (whose size is determined by underlying `bytearray`) is a temporary placeholder for chunks read from a file. Instead here we are reading the whole file in one shot for which `memoryview` intermediate step is not required at all. Also if the file size is large, this will try to keep the whole file in memory.

For now the easiest route will be to add a function `hash_file` in the underlying API interface, and move this same logic there, as in this way this function will have direct access to the file handle.

src/unstract/sdk/file_storage/fs_impl.py
gaya3-zipstack: For all such cases, the behaviour will be defined by the behaviour of the underlying fsspec lib @hari-kuriakose

src/unstract/sdk/file_storage/fs_impl.py
gaya3-zipstack: Silently ignored inside the implementation if file/path already exists

tests/sample.env
gaya3-zipstack: Ok @harini-venkataraman As I mentioned, this is only for unit tests cases that I have. So it should be ok to leave this unchanged.

gaya3-zipstack: Choosing the provider as discussed can be kept in core for easy use across services as anyway the file storage variants like Permanent and Temporary are kept in core.

src/unstract/sdk/utils/tool_utils.py
gaya3-zipstack: Yeah @hari-kuriakose . I explored a bit and looks like that is the best for now. I have addressed this. Pls take a re-look.
I have manually deprecated the older function keeping the backward compatibility.

src/unstract/sdk/adapters/ocr/google_document_ai/src/google_document_ai.py
```diff
@@ -95,10 +96,13 @@ def _get_request_headers(self) -> dict[str, Any]:
 
     """ Detect the mime type from the file content """
 
-    def _get_input_file_type_mime(self, input_file_path: str) -> str:
-        with open(input_file_path, mode="rb") as file_obj:
-            sample_contents = file_obj.read(100)
-            file_type = filetype.guess(sample_contents)
+    def _get_input_file_type_mime(
```
chandrasekharan-zipstack, line 99: NIT: @gaya3-zipstack please use the method from `/utils` instead which uses magic

src/unstract/sdk/file_storage/fs_impl.py
```diff
@@ -0,0 +1,275 @@
...
+            path (str): Path of the directory to be created
+            create_parents (bool): Specify if parent directories to be created
+            if any of the nested directory does not exist
+        """
+        try:
+            self.fs.mkdir(path=path, create_parents=create_parents)
+        except FileExistsError:
+            logger.debug(f"Path {path} already exists.")
+        except Exception as e:
+            raise FileOperationError(str(e))
```
chandrasekharan-zipstack, line 121: @gaya3-zipstack if we wish to wrap all exceptions from these implementations you can make use of a decorator to handle it and reduce nesting of code unecessarily

src/unstract/sdk/adapters/ocr/google_document_ai/src/google_document_ai.py
gaya3-zipstack: I am just refactoring existing function @chandrasekharan-zipstack I see more than one function for this and both of them being used. Will have to take the cleanup later.

tests/test_file_storage.py
```diff
@@ -0,0 +1,650 @@
...
+
+
+def file_storage(provider: FileStorageProvider):
+    try:
+        if provider == FileStorageProvider.GCS:
+            creds = json.loads(os.environ.get(TEST_CONSTANTS.FILE_STORAGE_GCS, {}))
+        elif provider == FileStorageProvider.Minio:
+            creds = json.loads(os.environ.get(TEST_CONSTANTS.FILE_STORAGE_MINIO, {}))
+        elif provider == FileStorageProvider.Local:
+            creds = json.loads(os.environ.get(TEST_CONSTANTS.FILE_STORAGE_LOCAL, {}))
```
kirtimanmishrazipstack, line 40: creds = json.loads(os.environ.get(TEST_CONSTANTS.FILE_STORAGE_LOCAL, ""))

tests/test_fs_permanent.py
```diff
@@ -0,0 +1,123 @@
...
+    FILE_STORAGE_GCS = "FILE_STORAGE_GCS"
+    FILE_STORAGE_LOCAL = "FILE_STORAGE_LOCAL"
+
+
+def permanent_file_storage(provider: FileStorageProvider):
+    try:
+        if provider == FileStorageProvider.GCS:
+            creds = json.loads(os.environ.get(TEST_CONSTANTS.FILE_STORAGE_GCS))
+        elif provider == FileStorageProvider.Local:
+            creds = json.loads(os.environ.get(TEST_CONSTANTS.FILE_STORAGE_LOCAL))
```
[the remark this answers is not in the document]
gaya3-zipstack: Yes this is optional. However in future if something is required, we can add. I have  some thoughts of keeping auto_mkdir also configurable in this list. Hence keeping this for now.

tests/test_file_storage.py
gaya3-zipstack: Is this the same  comment as above @kirtimanmishrazipstack ?

kirtimanmishrazipstack (changes requested):
creds = json.loads(os.environ.get(TEST_CONSTANTS.FILE_STORAGE_LOCAL, ""))

tests/test_file_storage.py
kirtimanmishrazipstack: creds = json.loads(os.environ.get(TEST_CONSTANTS.FILE_STORAGE_LOCAL, ""))

kirtimanmishrazipstack: @gaya3-zipstack can you do like this, else the test cases break

kirtimanmishrazipstack (changes requested):

tests/test_fs_permanent.py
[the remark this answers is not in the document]
kirtimanmishrazipstack: `creds = json.loads(os.environ.get(TEST_CONSTANTS.FILE_STORAGE_LOCAL, ""))`

can you make like this @gaya3-zipstack as it is optional

kirtimanmishrazipstack (changes requested):

tests/test_file_storage.py
kirtimanmishrazipstack: yes , can you make it like the comment above?

kirtimanmishrazipstack (approved):

src/unstract/sdk/file_storage/fs_permanent.py
```diff
@@ -45,8 +46,9 @@ def _copy_on_write(self, path):
         """
         if not self.exists(path):
             local_file_storage = FileStorage(provider=FileStorageProvider.Local)
-            if local_file_storage.exists(path):
-                self.fs.put(path, path)
+            local_file_path = self.legacy_storage_path + "/" + path
```
chandrasekharan-zipstack, line 49: NIT: I suggest using Pathlib to perform any path based manipulation rather than directly manipulating as strings

src/unstract/sdk/file_storage/helper.py
```diff
@@ -1,72 +1,58 @@
...
-            logger.debug(f"Connected to {FileStorageProvider.GCS.value} file system")
+            logger.debug(f"Connected to {provider.value} file system")
+        except KeyError as e:
+            logger.error(
+                f"Error in initialising {provider.value} "
+                f"file system because of missing config {e}",
+                stack_info=True,
+                exc_info=True,
+            )
+            raise FileStorageError(str(e))
```
chandrasekharan-zipstack, line 46: @gaya3-zipstack this will display the `KeyError`'s message only in the caller which might not be useful

Is there any reason why we've logged this error with a traceback also? I think as a library we needn't do this and the logging of the traceback can remain within the client (Unstract) unless we suppress an exception anywhere
cc: @hari-kuriakose

tests/test_file_storage.py
```diff
@@ -24,12 +25,22 @@ class TEST_CONSTANTS:
     TEST_FOLDER = os.environ.get("TEST_FOLDER")
     GCS_BUCKET = os.environ.get("GCS_BUCKET")
     TEXT_CONTENT = os.environ.get("TEXT_CONTENT")
-    FILE_STORAGE_ENV = "FILE_STORAGE"
+    FILE_STORAGE_GCS = "FILE_STORAGE_GCS"
+    FILE_STORAGE_MINIO = "FILE_STORAGE_MINIO"
+    FILE_STORAGE_LOCAL = "FILE_STORAGE_LOCAL"
 
 
 def file_storage(provider: FileStorageProvider):
```
chandrasekharan-zipstack, line 33: @gaya3-zipstack make this a pytest fixture with a valid scope that can be reused easily. You'd have to place it in a separate `conftest.py`

tests/test_file_storage.py
```diff
@@ -450,8 +461,10 @@ def test_seek_file(file_storage, file_path, mode, location, whence, expected_siz
 @pytest.mark.parametrize("provider", [(FileStorageProvider.GCS)])
```
chandrasekharan-zipstack, line 461: @gaya3-zipstack by using it as a fixture, you avoid the need to parameterize it and pass to your tests

src/unstract/sdk/adapters/ocr/google_document_ai/src/google_document_ai.py
chandrasekharan-zipstack: @gaya3-zipstack I suggested this for you to avoid the unnecessary work on this function - worst case, can you add a TODO comment to perform this cleanup later?

src/unstract/sdk/file_storage/constants.py
```diff
@@ -0,0 +1,9 @@
+class Common:
+    FULL = -1
+    DEFAULT_ENCODING = "utf-8"
```
chandrasekharan-zipstack, line 3:
```suggestion
    READ_ENTIRE_FILE = -1
    UTF_8 = "utf-8"
```
NIT: `Common` doesn't feel like a good name to define these constants. I suggest moving this within the `FileStorage` class where its actually associated / meant to be used. I've also suggested some alternate names to convey the use better

src/unstract/sdk/file_storage/fs_impl.py
```diff
@@ -0,0 +1,335 @@
...
+        try:
+            file_info = self.fs.info(path)
+            file_mtime = file_info["mtime"]
+            if not isinstance(file_mtime, datetime):
+                file_mtime = datetime.fromtimestamp(file_mtime)
+            return file_mtime
+        except Exception as e:
+            raise FileOperationError(str(e))
+
+    def mime_type(self, path: str) -> str:
```
chandrasekharan-zipstack, line 218: NIT: @gaya3-zipstack how is this different from the function in `/utils`? Can we aim to retain one of the implementations alone eventually?

src/unstract/sdk/file_storage/fs_permanent.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def __init__(
+        self,
+        provider: FileStorageProvider,
+        legacy_storage_path: Optional[str] = None,
+        **storage_config,
+    ):
+        if provider.value not in self.SUPPORTED_FILE_STORAGE_TYPES:
+            raise FileStorageError(
+                f"File storage provider is not supported in Permanent mode. "
```
chandrasekharan-zipstack, line 25:
```suggestion
                f"File storage provider `{provider.value}` is not supported in Permanent mode. "
```

src/unstract/sdk/file_storage/fs_shared_temporary.py
```diff
@@ -0,0 +1,26 @@
...
+class SharedTemporaryFileStorage(FileStorage):
+    SUPPORTED_FILE_STORAGE_TYPES = [
+        FileStorageProvider.Minio.value,
+        FileStorageProvider.Redis.value,
+    ]
+
+    def __init__(
+        self,
+        provider: FileStorageProvider,
+        storage_config,
```
chandrasekharan-zipstack, line 14:
```suggestion
        **storage_config
```
NIT: Might be good to add type hints wherever possible too

src/unstract/sdk/utils/tool_utils.py
```diff
@@ -47,16 +53,20 @@ def get_hash_from_file(file_path: str) -> str:
...
-        mv = memoryview(b)
-        with open(file_path, "rb", buffering=0) as f:
-            while n := f.readinto(mv):
-                h.update(mv[:n])
-        return str(h.hexdigest())
+
+        # Adding the following DeprecationWarning manually as the package "deprecated"
+        # does not support deprecation on static methods.
+        warnings.warn(
+            "`old_function` is deprecated. Use `new_function` instead.",
```
chandrasekharan-zipstack, line 60: NIT: Please specify the function names correctly - also it'll be good to mention the version it's deprecated from. Check if that's an argument to `warnings.warn()`

tests/test_fs_permanent.py
chandrasekharan-zipstack, line 1: Can these tests be automated to run on every change we push?

src/unstract/sdk/file_storage/fs_impl.py
```diff
@@ -227,10 +246,35 @@ def download(self, from_path: str, to_path: str):
...
+        Returns:
+            NA
+        """
+        try:
+            self.fs.put(from_path, to_path)
+        except FileNotFoundError as e:
+            # if to_path in str(e):
+            #     self.fs.touch(to_path)
+            #     logger.info(f"Creating file structure {to_path}")
+            # else:
```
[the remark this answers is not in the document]
gaya3-zipstack: Removed the commented lines.

src/unstract/sdk/file_storage/helper.py
gaya3-zipstack: Yeah... I have removed this now. Caller can log the details...

src/unstract/sdk/adapters/ocr/google_document_ai/src/google_document_ai.py
```diff
@@ -110,13 +114,15 @@ def _get_input_file_type_mime(self, input_file_path: str) -> str:
...
+        input_file_path: str,
+        output_file_path: Optional[str] = None,
+        fs: FileStorage = FileStorage(provider=FileStorageProvider.Local),
     ) -> str:
         try:
             file_type_mime = self._get_input_file_type_mime(input_file_path)
-            if os.path.isfile(input_file_path):
-                with open(input_file_path, "rb") as fop:
-                    file_content_in_bytes: bytes = fop.read()
+            if fs.exists(input_file_path):
```
muhammad-ali-e, line 124: is this `fs.exists` functioning same function of isfile? I think `isfile` returns False for Directory, hope the same behaviour in `exists`

src/unstract/sdk/file_storage/fs_impl.py
gaya3-zipstack: Idea is to retain one.  Once we have moved everywhere to FileStorage, we should have only this one and remove the other versions.

gaya3-zipstack: I have added ToDO in other functions.

tests/test_fs_permanent.py
gaya3-zipstack: Yes we can take it up later...

src/unstract/sdk/adapters/utils.py
```diff
@@ -45,15 +49,15 @@ def get_file_mime_type(input_file: Path) -> str:
         Returns:
             str: MIME type of the file
         """
-        input_file_mime = ""
-        with open(input_file, mode="rb") as input_file_obj:
-            sample_contents = input_file_obj.read(100)
-            input_file_mime = magic.from_buffer(sample_contents, mime=True)
-            input_file_obj.seek(0)
+        sample_contents = fs.read(path=input_file, mode="rb", length=100)
```
muhammad-ali-e, line 52: Increasing the default `length` to a higher value, such as `1024`, would optimize the reading process by reducing the number of iterations.

src/unstract/sdk/utils/tool_utils.py
gaya3-zipstack: Thank you for spotting this. Done.

src/unstract/sdk/adapters/x2text/llm_whisperer/src/llm_whisperer.py
```diff
@@ -360,22 +371,36 @@ def _write_output_to_file(self, output_json: dict, output_file_path: Path) -> No
...
             text_output = output_json.get("text", "")
             logger.info(f"Writing output to {output_file_path}")
-            output_file_path.write_text(text_output, encoding="utf-8")
+            fs.write(
+                path=output_file_path.name,
+                mode="w",
+                encoding="utf-8",
+                data=text_output,
+            )
+            # output_file_path.write_text(text_output, encoding="utf-8")
```
muhammad-ali-e, line 380: remove this commented lines if not required

src/unstract/sdk/adapters/x2text/llm_whisperer/src/llm_whisperer.py
```diff
@@ -360,22 +371,36 @@ def _write_output_to_file(self, output_json: dict, output_file_path: Path) -> No
...
             try:
                 # Define the directory of the output file and metadata paths
                 output_dir = output_file_path.parent
                 metadata_dir = output_dir / "metadata"
                 metadata_file_name = output_file_path.with_suffix(".json").name
                 metadata_file_path = metadata_dir / metadata_file_name
                 # Ensure the metadata directory exists
-                metadata_dir.mkdir(parents=True, exist_ok=True)
+                fs.mkdir(metadata_dir, create_parents=True)
+                # metadata_dir.mkdir(parents=True, exist_ok=True)
```
muhammad-ali-e, line 389: remove this commented lines if not required

src/unstract/sdk/adapters/x2text/llm_whisperer/src/llm_whisperer.py
```diff
@@ -360,22 +371,36 @@ def _write_output_to_file(self, output_json: dict, output_file_path: Path) -> No
...
                 # Remove the "text" key from the metadata
                 metadata = {
                     key: value for key, value in output_json.items() if key != "text"
                 }
                 metadata_json = json.dumps(metadata, ensure_ascii=False, indent=4)
                 logger.info(f"Writing metadata to {metadata_file_path}")
-                metadata_file_path.write_text(metadata_json, encoding="utf-8")
+                # metadata_file_path.write_text(metadata_json, encoding="utf-8")
```
muhammad-ali-e, line 396: remove this commented lines if not required

src/unstract/sdk/file_storage/fs_permanent.py
gaya3-zipstack: I agree. However, I do not see many use cases where we do this for now. Also this is a piece which can be removed later as this is only to support legacy migration for now.... Hence leaving it for now. Hope that is ok.

src/unstract/sdk/exceptions.py
```diff
@@ -39,3 +39,11 @@ class OCRError(SdkError):
 
 class RateLimitError(SdkError):
     DEFAULT_MESSAGE = "Running into rate limit errors, please try again later"
+
+
+class FileStorageError(SdkError):
+    pass
```
muhammad-ali-e, line 45: Adding a default message here would be beter to avoid it return `Somwthing went wrong`

src/unstract/sdk/file_storage/constants.py
muhammad-ali-e: `Full` is might be  better to change to `READ_ENTIRE_FILE`. but `UTF_8` not seems a  better replaceble for  DEFAULT_ENCODING. If we need UTF_8 we should create a new class for constants like ENCODINGS .

src/unstract/sdk/file_storage/fs_shared_temporary.py
gaya3-zipstack: Done.

src/unstract/sdk/adapters/ocr/google_document_ai/src/google_document_ai.py
gaya3-zipstack: Currently, exists will only indicate if the object path exists.

src/unstract/sdk/adapters/utils.py
gaya3-zipstack: @muhammad-ali-e the idea is for this function to go away post file storage APIs are integrated and tested. Would prefer to not change this as this will be deleted in some time.

src/unstract/sdk/adapters/x2text/llm_whisperer/src/llm_whisperer.py
gaya3-zipstack: Done

src/unstract/sdk/adapters/x2text/llm_whisperer/src/llm_whisperer.py
gaya3-zipstack: Done

src/unstract/sdk/adapters/x2text/llm_whisperer/src/llm_whisperer.py
gaya3-zipstack: Done

src/unstract/sdk/exceptions.py
gaya3-zipstack: Done

src/unstract/sdk/file_storage/constants.py
gaya3-zipstack: Done

src/unstract/sdk/adapters/ocr/google_document_ai/src/google_document_ai.py
gaya3-zipstack: Done.

src/unstract/sdk/file_storage/fs_impl.py
gaya3-zipstack: @chandrasekharan-zipstack  The challenge that I see here is that fsspec does not have a documented set of exceptions so that they can be handled explicitly. There are references indicating that they transmit the underlying exception from the inetgration points -  I think in such cases, forwarding the exceptions as is will be the best for now.  Could you pls point what sort of decorators are you talking about?

chandrasekharan-zipstack: @gaya3-zipstack my main concern was around catching all exceptions and not specific exceptions. Let's see how this pans out for us and we can change it later if required.
The decorator I suggested was something generic such as [this](https://github.com/Zipstack/unstract-sdk/blob/4b44c02ef9b26f28adc3ad1d3053bea39120f6a6/src/unstract/sdk/utils/common_utils.py#L34-L55) where we perform this catch and re-raise

src/unstract/sdk/utils/tool_utils.py
```diff
@@ -93,6 +94,8 @@ def json_to_str(json_to_dump: dict[str, Any]) -> str:
         compact_json = json.dumps(json_to_dump, separators=(",", ":"))
         return compact_json
 
+    # ToDo: get_file_mime_type() to be removed once migrated to FileStorage
```
chandrasekharan-zipstack, line 97:
```suggestion
    # TODO: get_file_mime_type() to be removed once migrated to FileStorage
```

tests/test_fs_permanent.py
chandrasekharan-zipstack: @gaya3-zipstack I request you to create a ticket for this maybe - I don't think we'll leverage the true power of these tests that you've invested your time in unless we automate them to catch regressions. Feel free to resolve this conversation afterwards

chandrasekharan-zipstack (approved):
LGTM overall, left some minor comments alone

src/unstract/sdk/file_storage/fs_provider.py
```diff
@@ -0,0 +1,10 @@
+import enum
+
+
+class FileStorageProvider(enum.Enum):
+    Azure = "azure"
```
muhammad-ali-e, line 5: UPPER_CASE is prefered for enum names.  would be better to validate the enum by to upper.

ref: https://docs.python.org/3/howto/enum.html#:~:text=Note%20Case%20of,in%20our%20examples.

harini-venkataraman (approved):

harini-venkataraman (approved):

muhammad-ali-e (approved):
Left a comment regaurding the enum style

src/unstract/sdk/exceptions.py
```diff
@@ -39,3 +39,18 @@ class OCRError(SdkError):
...
+        "Please check the configuration credentials"
+    )
+
+
+class FileOperationError(SdkError):
+    DEFAULT_MESSAGE = (
+        "Error while performing operation on the file. "
+        "Please check specific storage error for "
+        "further information"
+    )
```
chandrasekharan-zipstack, line 56: @gaya3-zipstack is there any reason why we need to differentiate both these types of errors? How do you think it might help downstream usecases?
We could try to have a common subclass of `SdkError` that both these classes can inherit from to simplify the error handling maybe

src/unstract/sdk/file_storage/fs_shared_temporary.py
```diff
@@ -0,0 +1,28 @@
...
+        provider: FileStorageProvider,
+        **storage_config: dict[str, Any],
+    ):
+        if provider.value not in self.SUPPORTED_FILE_STORAGE_TYPES:
+            raise FileStorageError(
+                f"File storage provider is not supported in Permanent mode. "
+                f"Supported providers: {self.SUPPORTED_FILE_STORAGE_TYPES}"
+            )
+        if provider == FileStorageProvider.Minio:
+            super().__init__(provider, storage_config)
```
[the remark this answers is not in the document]
gaya3-zipstack: Thank you @muhammad-ali-e . Done.

src/unstract/sdk/exceptions.py
gaya3-zipstack: FileOperationError happens when we try to work on a file
FileStorageError happens when we try to get the storage connection.
Common exception - Do you  mean that this will help handling at the caller's side @chandrasekharan-zipstack ?

src/unstract/sdk/file_storage/fs_provider.py
gaya3-zipstack: Done.

kirtimanmishrazipstack (approved):

harini-venkataraman (approved):

src/unstract/sdk/file_storage/fs_impl.py
gaya3-zipstack: Ok will look into this once the base is out and working.

src/unstract/sdk/exceptions.py
gaya3-zipstack: Will take up any improvements incrementally

tests/test_fs_permanent.py
gaya3-zipstack: FOr now it is useful for me to test before I commit anything. But yes, we need an overall plan to autimate all such unit tests. - https://zipstack.atlassian.net/browse/UN-1763

tests/test_file_storage.py
gaya3-zipstack: Like I mentioned, I have  a use case where I need to support arguments to this function which er are trying to keep as fixture. I could not find a way to do that.

tests/test_file_storage.py
gaya3-zipstack: Will take it up later

src/unstract/sdk/file_storage/helper.py
```diff
@@ -0,0 +1,55 @@
...
+from unstract.sdk.exceptions import FileStorageError
+from unstract.sdk.file_storage.constants import GCS
+from unstract.sdk.file_storage.fs_provider import FileStorageProvider
+
+logger = logging.getLogger(__name__)
+
+
+class FileStorageHelper:
+    @staticmethod
+    def file_storage_init(provider: FileStorageProvider, credentials):
```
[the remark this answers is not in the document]
gaya3-zipstack: Done.

src/unstract/sdk/file_storage/fs_impl.py
gaya3-zipstack: Taking up later.

muhammad-ali-e (approved):
LGTM

src/unstract/sdk/file_storage/fs_impl.py
gaya3-zipstack: This logic has changed now.

kirtimanmishrazipstack (approved):

tests/test_file_storage.py
gaya3-zipstack: Comment is outdated as APIs have changed.

PR #120, opened 2024-11-15
feat: Enhanced error handling to clearly point third party errors

chandrasekharan-zipstack:
NOTE: This requires some corresponding FE changes as well which will be a part of Unstract
## What

- Refactored `parse_llm_err`
- Handled few error scenarios for anthropic
- Changes to SDK's error handling to include status code and actual error that occurred
- Added `parse_vector_db_error` which **needs to be** implemented for other providers based on need
- Updated a few JSON schema to include URLs in the help text

## Why

- Confusion arises in case of 3rd party errors and there's a need to clearly point it out
- We incorrectly raise 3rd party errors with 5xx status codes which might cause error fatigue while monitoring

## How

- Using few markdown components to render URLs / codeblocks

## Important files to review
- [LLM error parsing](https://github.com/Zipstack/unstract-sdk/pull/120/files#diff-e1ab636dc8b2965981bba66af53ee7d2083e9e5b91fd1f57bd5cb2fb57107d6e)
- [vector DB error parsing](https://github.com/Zipstack/unstract-sdk/pull/120/files#diff-c0f0653bd26f8c7dad0f31dd44b04ce510e5937156d1c2a7bf22720a42acce4e)
- [SDK error parsing](https://github.com/Zipstack/unstract-sdk/pull/120/files#diff-2aa004a08bcea12c0e185e792e2591bf93de2c6e934d4165b1bae1b252703018)
- [Index related changes](https://github.com/Zipstack/unstract-sdk/pull/120/files#diff-47fd4f8fea94f6b5d2258eff13143216acd67eea8ddc713e799693c958c1a227)

## Notes on Testing

- Tried some error cases around testing of adapters and its use in prompt studio / workflow
- Couldn't simulate for API deployment yet

## Screenshots
1. Error while testing qdrant
image
1. Links added to few forms
image
image
1. Error handling in prompt studio
image

## Checklist


hari-kuriakose (approved):
@chandrasekharan-zipstack Looks good overall.

src/unstract/sdk/adapters/exceptions.py
```diff
@@ -9,13 +10,5 @@ class LLMError(AdapterError):
     pass
 
 
-class VectorDBError(AdapterError):
```
gaya3-zipstack, line 12: Did we remove this because it is not being used anywhere?

src/unstract/sdk/adapters/vectordb/exceptions.py
```diff
@@ -0,0 +1,37 @@
+from qdrant_client.http.exceptions import ApiException as QdrantAPIException
+
+from unstract.sdk.adapters.vectordb.qdrant.src import Qdrant
+from unstract.sdk.adapters.vectordb.vectordb_adapter import VectorDBAdapter
+from unstract.sdk.exceptions import VectorDBError
```
gaya3-zipstack, line 5: Is VectorDBError defined in exceptions? I see that it is removed in this PR(?)

tests/test_fs_permanent.py
```diff
@@ -5,8 +5,7 @@
 import pytest
 from dotenv import load_dotenv
 
-from unstract.sdk.file_storage import FileStorageProvider
-from unstract.sdk.file_storage.fs_permanent import PermanentFileStorage
+from unstract.sdk.file_storage import FileStorageProvider, PermanentFileStorage
```
gaya3-zipstack, line 8: My PR already is having these changes as suggested by you.

src/unstract/sdk/adapters/exceptions.py
gaya3-zipstack: Ok, I see these being defined in unstract.sdk.exceptions. Should LLMError also be removed from here then?

src/unstract/sdk/llm.py
```diff
@@ -109,7 +109,7 @@ def complete(
                     process_text_output = {}
             return {LLM.RESPONSE: response, **process_text_output}
         except Exception as e:
-            raise parse_llm_err(e) from e
+            raise parse_llm_err(e, self._llm_instance) from e
```
gaya3-zipstack, line 112: Wondering if we can have an approach where we do
self.parse_error()  # self will point to an Unstract LLM Adapter instance for example
-> Anthropic.parse_error()
   -> here we parse Anthropic specific errors alone

The approach we have taken now is to have parse functions at the LLM level where we check for the exception type and then route to individual Unstract Adapter type's parse() functions.

Instead if we use botton up approach where we define parse functions at the Unstract adapter type's implementation and make sure that based on the llm instance used, the call will land to the specific parse() ?

We can have a discussion around this if the msg is not clear.

src/unstract/sdk/adapters/llm/exceptions.py
```diff
@@ -24,14 +25,29 @@ def parse_llm_err(e: Exception) -> LLMError:
...
     elif isinstance(e, OpenAIAPIError):
-        return OpenAILLM.parse_llm_err(e)
+        err = OpenAILLM.parse_llm_err(e)
     elif isinstance(e, AnthropicAPIError):
-        return AnthropicLLM.parse_llm_err(e)
+        err = AnthropicLLM.parse_llm_err(e)
     elif isinstance(e, MistralException):
-        return MistralLLM.parse_llm_err(e)
+        err = MistralLLM.parse_llm_err(e)
     elif isinstance(e, GoogleAPICallError):
```
gaya3-zipstack, line 40: There could be cases where GoogleAPICallError is thrown for Gemini/VertexAI also.... Instead we should allow the the context (in this case, the llm instance) itself to take us to the leaf implementation of the parse()

src/unstract/sdk/adapters/exceptions.py
chandrasekharan-zipstack: @gaya3-zipstack yes, to avoid confusion and to keep it simple we should aim to remove the other exceptions defined in this file. I've added a TODO for it since I noticed some usages. I'll try to remove those as well but I think its being used downstream as well so maybe we could plan it later too

src/unstract/sdk/adapters/vectordb/exceptions.py
chandrasekharan-zipstack: Yes, its always been present and it inherits from `SdkError`

tests/test_fs_permanent.py
chandrasekharan-zipstack: Will resolve any conflicts if it comes up

src/unstract/sdk/adapters/llm/exceptions.py
chandrasekharan-zipstack: Answered [here](https://github.com/Zipstack/unstract-sdk/pull/120#discussion_r1869046536)

PR #134, opened 2024-12-06
Support length to be passed while identifying mime type

gaya3-zipstack:
## What

Supporting passing in the number of bytes to be read while reading a file that is passed to understand the file's mime type

## Why

For certain file types, the full length needs to be read to get the mime type correctly. Supporting this as an optional param allows the user to use this if at all required based on the use case.

## How

Using an optional param. This is defaulted to 100 bytes.

## Relevant Docs

-

## Related Issues or PRs

-

-

## Notes on Testing

All unit test cases ran and  added case to test the new additon.

## Screenshots

...

## Checklist

### DIFF
--- a/src/unstract/sdk/file_storage/env_helper.py
+++ b/src/unstract/sdk/file_storage/env_helper.py
@@ -13,8 +13,24 @@ logger = logging.getLogger(__name__)
 
 
 class EnvHelper:
+    ENV_CONFIG_FORMAT = (
+        '{"provider": "gcs", ' '"credentials": {"token": "/path/to/google/creds.json"}}'
+    )
+
     @staticmethod
     def get_storage(storage_type: StorageType, env_name: str) -> FileStorage:
+        """Helper function for clients to pick up remote storage configuration
+        from env, initialise the file storage for the same and return the
+        instance.
+
+        Args:
+            storage_type: Permanent / Temporary file storage
+            env_name: Name of the env which has the file storage config
+
+        Returns:
+            FileStorage: FIleStorage instance initialised using the provider
+            and credentials configured in the env
+        """
         try:
             file_storage_creds = json.loads(os.environ.get(env_name))
             provider = FileStorageProvider(
@@ -31,7 +47,8 @@ class EnvHelper:
                 raise NotImplementedError()
             return file_storage
         except KeyError as e:
-            logger.error(f"Required credentials is missing in the env: {str(e)}")
+            logger.error(f"Required credentials are missing in the env: {str(e)}")
+            logger.error(f"The configuration format is {EnvHelper.ENV_CONFIG_FORMAT}")
             raise e
         except FileStorageError as e:
             raise e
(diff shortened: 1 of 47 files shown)


chandrasekharan-zipstack (approved):
LGTM - this will be in `rc6` btw

harini-venkataraman (approved):
LGTM

harini-venkataraman (approved):

src/unstract/sdk/file_storage/env_helper.py
```diff
@@ -32,7 +32,7 @@ def get_storage(storage_type: StorageType, env_name: str) -> FileStorage:
             and credentials configured in the env
         """
         try:
-            file_storage_creds = json.loads(os.environ.get(env_name))
+            file_storage_creds = json.loads(os.environ.get(env_name, ""))
```
muhammad-ali-e, line 35: I think below code should need instaed of default empty string
```
file_storage_creds = json.loads(os.environ.get(env_name, "{}"))
```
But anyway, This default value might not help here. If it getting empty or None here the next line `37` will raise key error. right? So it would be better if we handle such exception here and raise meaning full error message or meaning full exception.

src/unstract/sdk/file_storage/permanent.py
```diff
@@ -28,10 +28,8 @@ def __init__(
                 f"supported in Permanent mode. "
                 f"Supported providers: {self.SUPPORTED_FILE_STORAGE_TYPES}"
             )
-        if provider == FileStorageProvider.GCS:
+        if provider == FileStorageProvider.GCS or provider == FileStorageProvider.LOCAL:
             super().__init__(provider, **storage_config)
-        elif provider == FileStorageProvider.LOCAL:
-            super().__init__(provider)
         else:
             raise NotImplementedError
```
muhammad-ali-e, line 34: How this `NotImplementedError` diffr from `FileStorageError`. For exmple for Azure they will not get `FileStorageError` but `NotImplementedError`.

And `NotImplementedError` doesn't return what are we implemented or suported providers.

muhammad-ali-e (approved):
Added  minor comments, Aprt from that LGTM

src/unstract/sdk/file_storage/env_helper.py
gaya3-zipstack: Thank you. Changed.
Actually the exception is caught and an error message is displayed with the right format based in your feedback in the last PR. Pls look for it in lines 49-51 below.

src/unstract/sdk/file_storage/permanent.py
gaya3-zipstack: NotImplemented is raised when we want to mean that though it is supported as a Permanent mode, we have not implemented it yet. Some conntributir can write the impl for Azure and then escape it from this error. However, FileStorageError in this context would mean that we do no intend to support if for PermanentFIleStorage at all.

gaya3-zipstack: I have now added an error message also. Hope that helps.

muhammad-ali-e (approved):

PR #136, opened 2024-12-17
[FEATURE] Capture metrics

Deepak-Kesavan:
## What

- Added `MetricsMixin` to capture time taken in each step
- Added `capture_metrics` decorator function

## Why

- To capture metrics and return the time taken

## How

...

## Relevant Docs

-

## Related Issues or PRs

https://github.com/Zipstack/unstract/pull/900

-

## Notes on Testing

...

## Screenshots

...

## Checklist

### DIFF
--- a/src/unstract/sdk/llm.py
+++ b/src/unstract/sdk/llm.py
@@ -19,6 +19,7 @@ from unstract.sdk.exceptions import LLMError, RateLimitError, SdkError
 from unstract.sdk.helper import SdkHelper
 from unstract.sdk.tool.base import BaseTool
 from unstract.sdk.utils.callback_manager import CallbackManager
+from unstract.sdk.utils.common_utils import capture_metrics
 
 logger = logging.getLogger(__name__)
 
@@ -36,6 +37,7 @@ class LLM:
         tool: BaseTool,
         adapter_instance_id: Optional[str] = None,
         usage_kwargs: dict[Any, Any] = {},
+        capture_metrics: bool = False,
     ):
         """Creates an instance of this LLM class.
 
@@ -50,6 +52,10 @@ class LLM:
         self._adapter_instance_id = adapter_instance_id
         self._llm_instance: LlamaIndexLLM = None
         self._usage_kwargs = usage_kwargs
+        self._capture_metrics = capture_metrics
+        self._run_id = usage_kwargs.get("run_id")
+        self._usage_reason = usage_kwargs.get("llm_usage_reason")
+        self._metrics = {}
         self._initialise()
 
     def _initialise(self):
@@ -65,6 +71,7 @@ class LLM:
                     kwargs=self._usage_kwargs,
                 )
 
+    @capture_metrics
     def complete(
         self,
         prompt: str,
@@ -72,7 +79,8 @@ class LLM:
         process_text: Optional[Callable[[str], str]] = None,
         **kwargs: Any,
     ) -> dict[str, Any]:
-        """Generates a completion response for the given prompt.
+        """Generates a completion response for the given prompt and captures
+        metrics if run_id is provided.
 
         Args:
             prompt (str): The input text prompt for generating the completion.
@@ -85,12 +93,8 @@ class LLM:
             **kwargs (Any): Additional arguments passed to the completion function.
 
         Returns:
-            dict[str, Any]: A dictionary containing the result of the completion
-                and any processed output.
-
-        Raises:
-            LLMError: If an error occurs during the completion process, it will be
-                raised after being processed by `parse_llm_err`.
+            dict[str, Any]: A dictionary containing the result of the completion,
+                any processed output, and the captured metrics (if applicable).
         """
         try:
             response: CompletionResponse = self._llm_instance.complete(prompt, **kwargs)
@@ -105,11 +109,18 @@ class LLM:
                     if not isinstance(process_text_output, dict):
                         process_text_output = {}
                 except Exception as e:
-                    logger.error(f"Error occured inside function 'process_text': {e}")
+                    logger.error(f"Error occurred inside function 'process_text': {e}")
                     process_text_output = {}
-            return {LLM.RESPONSE: response, **process_text_output}
+            response_data = {LLM.RESPONSE: response, **process_text_output}
+            return response_data
         except Exception as e:
             raise parse_llm_err(e, self._llm_instance) from e
+
+    def get_metrics(self):
+        return self._metrics
+
+    def get_usage_reason(self):
+        return self._usage_reason
 
     def stream_complete(
         self,
--- a/src/unstract/sdk/metrics_mixin.py
+++ b/src/unstract/sdk/metrics_mixin.py
@@ -44,12 +44,17 @@ class MetricsMixin:
         Returns:
             dict: The calculated time taken and the associated run_id and op_id.
         """
+        # Check if the key exists in Redis
+        if not self.redis_client.exists(self.redis_key):
+            # If the key is missing, return an empty metrics dictionary
+            return {self.TIME_TAKEN_KEY: None}
+
         start_time = float(
             self.redis_client.get(self.redis_key)
         )  # Get the stored timestamp
-        time_taken = round(time.time() - float(start_time), 3)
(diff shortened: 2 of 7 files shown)


src/unstract/sdk/metrics_mixin.py
```diff
@@ -0,0 +1,51 @@
...
+
+class MetricsMixin:
+    def __init__(self, run_id):
+        """Initialize the MetricsMixin class.
+
+        Args:
+            run_id (str): Unique identifier for the run.
+        """
+        self.run_id = run_id
+        self.sub_id = str(uuid.uuid4())  # Unique identifier for this instance
```
chandrasekharan-zipstack, line 16: NIT: Could be named `metric_id` / `op_id` (for operation)

src/unstract/sdk/metrics_mixin.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+class MetricsMixin:
+    def __init__(self, run_id):
+        """Initialize the MetricsMixin class.
+
+        Args:
+            run_id (str): Unique identifier for the run.
+        """
+        self.run_id = run_id
```
chandrasekharan-zipstack, line 15: @Deepak-Kesavan can you confirm if this `run_id` will be unique for an entire tool run / file being executed? Where is it generated from? From [here](https://github.com/Zipstack/unstract/blob/3c491e51a07a1053f73687e2c05dfa72c9c1ebc5/backend/workflow_manager/workflow_v2/workflow_helper.py#L246)?

src/unstract/sdk/metrics_mixin.py
```diff
@@ -0,0 +1,51 @@
...
+        """Calculate the time taken since the timestamp was set and delete the
+        Redis key.
+
+        Returns:
+            dict: The calculated time taken and the associated run_id and sub_id.
+        """
+        start_time = float(
+            self.redis_client.get(self.redis_key)
+        )  # Get the stored timestamp
+        time_taken = round(time.time() - float(start_time), 2)
```
chandrasekharan-zipstack, line 46: NIT: Might be good to include upto 3 decimal places maybe if the metrics can get granular

src/unstract/sdk/utils/common_utils.py
```diff
@@ -54,3 +55,35 @@ def wrapper(*args, **kwargs):
...
     return decorator
+
+
+def capture_metrics(func):
+    """Decorator to capture metrics at the start and end of a function."""
+
+    @functools.wraps(func)
+    def wrapper(self, *args, **kwargs):
+        metrics_mixin = None
+        TIME_TAKEN_KEY = "time_taken(s)"
```
chandrasekharan-zipstack, line 66: NIT: Constant can be moved to `MetricsMixin` class or module and referred from there. Will be a single place to change / update this

src/unstract/sdk/metrics_mixin.py
```diff
@@ -0,0 +1,51 @@
...
+
+        # Set the start time immediately upon initialization
+        self.set_start_time()
+
+    def set_start_time(self, ttl=86400):
+        """Store the current timestamp in Redis when the instance is
+        created."""
+        self.redis_client.set(self.redis_key, time.time(), ex=ttl)
+
+    def collect_metrics(self):
```
chandrasekharan-zipstack, line 36:
```suggestion
    def collect_metrics(self) -> dict[str, Any]:
```
Please add relevant imports if you're approving this suggestion

src/unstract/sdk/utils/common_utils.py
```diff
@@ -54,3 +55,35 @@ def wrapper(*args, **kwargs):
         return wrapper
 
     return decorator
+
+
+def capture_metrics(func):
+    """Decorator to capture metrics at the start and end of a function."""
+
+    @functools.wraps(func)
+    def wrapper(self, *args, **kwargs):
```
chandrasekharan-zipstack, line 64: NIT: It'll be safer to check if `self` contains the attributes `_run_id` / `_capture_metrics` and `_metrics` beforehand to avoid any errors. If this decorator is used in a class which doesn't have these attributes, it can result in an error during the collection process

chandrasekharan-zipstack (approved):
LGTM - here are few suggestions for later
1. Try using a base interface to inherit / implement necessary metrics capturing attributes and functions. With the current logic it might not be easy to extend to other parts of the SDK in a robust fashion
1. Usage reason / `self.__class__.__name__` can be a part of this metrics dict returned by the mixin class

chandrasekharan-zipstack:
@Deepak-Kesavan can you also ensure that errors during the metrics collection are suppressed and logged as a warning to not prevent the user from accessing something?

src/unstract/sdk/utils/common_utils.py
```diff
@@ -54,3 +55,51 @@ def wrapper(*args, **kwargs):
...
+                    and time_taken_key in new_metrics
+                ):
+                    time_taken_self = self._metrics.get(time_taken_key)
+                    time_taken_new = new_metrics.get(time_taken_key)
+
+                    # Only sum if both are valid
+                    if time_taken_self and time_taken_new:
+                        self._metrics[time_taken_key] = time_taken_self + time_taken_new
+                    else:
+                        self._metrics[time_taken_key] = None
```
chandrasekharan-zipstack, line 98: NIT: Currently this might be fine - but @Deepak-Kesavan do you think its okay to set this to `None` if one of the `time_taken_*` is not available?

src/unstract/sdk/utils/common_utils.py
```diff
@@ -54,3 +55,51 @@ def wrapper(*args, **kwargs):
...
+                new_metrics = metrics_mixin.collect_metrics()
+
+                # If time_taken(s) exists in both self._metrics and new_metrics, sum it
+                if (
+                    self._metrics
+                    and time_taken_key in self._metrics
+                    and time_taken_key in new_metrics
+                ):
+                    time_taken_self = self._metrics.get(time_taken_key)
+                    time_taken_new = new_metrics.get(time_taken_key)
```
chandrasekharan-zipstack, line 92: NIT: Suggest to rename these as `previously_measured_time` and `newly_measured_time` maybe

src/unstract/sdk/metrics_mixin.py
chandrasekharan-zipstack: Discussed this offline - we need to ensure that the `run_id` mentioned in the link above needs to be passed down to the tools and used while reporting metrics

chandrasekharan-zipstack (approved):
LGTM overall. @Deepak-Kesavan you might want to catch any exceptions coming from the wrapper / MetricsMixin and log them as warnings to not block a user from executing something

src/unstract/sdk/utils/common_utils.py
Deepak-Kesavan: Yes @chandrasekharan-zipstack . I added it as fallback if key is not available in redis. Let's see how this works and I believe it will never hit this case.

src/unstract/sdk/llm.py
```diff
@@ -50,6 +52,10 @@ def __init__(
         self._adapter_instance_id = adapter_instance_id
         self._llm_instance: LlamaIndexLLM = None
         self._usage_kwargs = usage_kwargs
+        self._capture_metrics = capture_metrics
```
gaya3-zipstack, line 55: Are we really using _capture_metrics anywhere? If not, we could remove this var...

gaya3-zipstack: Is this an attempt to follow some guideline to keep the instance variables cached for fututr use? Because I see other constructor vars also being stored below.

Deepak-Kesavan: @gaya3-zipstack  It is being used in the `capture_metrics` decorator.

gaya3-zipstack (approved):
Changes  look fine.

src/unstract/sdk/metrics_mixin.py
Deepak-Kesavan: Yes @chandrasekharan-zipstack . I believe it is not being done like that currently. Need to handle it in unstract side.

PR #137, opened 2024-12-18
Fix enum conditional for env parsing

gaya3-zipstack:
## What

Fix a minor bug when matching enum values in env helper of file storage

## Why

Either enum's values need to be compared or the enum object needs to be compared. COde has a bug where on LHS it is enum object and on RHS it is enum value.

## How

By changing the data type on both sides

## Relevant Docs

-

## Related Issues or PRs

-

-

## Notes on Testing

Changed unit test case which was written wrongly and the bug went unntoiced.

## Screenshots

...

## Checklist


Deepak-Kesavan (approved):

harini-venkataraman (approved):
LGTM

PR #140, opened 2025-01-04
fixed connection closing prematurely in index.py

pk-zipstack:
## What

Fix for premature connection closure in vector database queries affecting Qdrant and Weaviate integrations.

## Why

The vector database connection was being closed too early due to a misplaced `finally` block, causing query attempts to fail with "Cannot send a request, as the client has been closed" errors. This was affecting prompt execution in both prompt studio.

## How

- Removed the redundant `finally` block that was closing the connection prematurely
- Consolidated error handling into a single try-except-finally block
- Ensured the vector database connection remains open throughout the query execution
- Moved connection cleanup to a single `finally` block at the end of the function

All the above changes are made in index.py

## Relevant Docs

-

## Related Issues or PRs

-

-

## Notes on Testing

...

## Screenshots

...

## Checklist

### DIFF
--- a/src/unstract/sdk/index.py
+++ b/src/unstract/sdk/index.py
@@ -89,17 +89,8 @@ class Index:
                 filters=filters,
                 similarity_top_k=Constants.TOP_K,
             )
-        except Exception as e:
-            self.tool.stream_log(
-                f"Error building query {vector_db}: {e}", level=LogLevel.ERROR
-            )
-            raise VectorDBError(
-                f"Error building query for {vector_db}: {e}", actual_err=e
-            )
-        finally:
-            vector_db.close()
 
-        try:
+            # Execute query with the open connection
             n: VectorStoreQueryResult = vector_db.query(query=q)
             if len(n.nodes) > 0:
                 self.tool.stream_log(f"Found {len(n.nodes)} nodes for {doc_id}")
@@ -110,7 +101,16 @@ class Index:
             else:
                 self.tool.stream_log(f"No nodes found for {doc_id}")
                 return None
+
+        except Exception as e:
+            self.tool.stream_log(
+                f"Error building query {vector_db}: {e}", level=LogLevel.ERROR
+            )
+            raise VectorDBError(
+                f"Error building query for {vector_db}: {e}", actual_err=e
+            )
         finally:
+            # Close connection only once at the end
             vector_db.close()
 
     @log_elapsed(operation="EXTRACTION")
--- a/src/unstract/sdk/index.py
+++ b/src/unstract/sdk/index.py
@@ -72,45 +72,50 @@ class Index:
         )
 
         try:
-            self.tool.stream_log(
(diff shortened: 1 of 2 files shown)


src/unstract/sdk/index.py
```diff
@@ -110,7 +101,16 @@ def query_index(
             else:
                 self.tool.stream_log(f"No nodes found for {doc_id}")
                 return None
+
+        except Exception as e:
+            self.tool.stream_log(
```
gaya3-zipstack, line 106: There are 2 operations happening here.
1. building query
2. actual query
Exceptions may be arising from either. If we have one try-except, then we need to change the log message appropriately.
Otherwie, we could have 2 try-except blocks enclosing the individual operations indicating the actual cause of the exception in the log message. We need to accordingly handle vector db close correctly.

pk-zipstack: I have separated query building and query execution into distinct try-except blocks again with a parent try-except block enclosing them. Would appreciate your feedback on whether this structure better addresses the concerns raised. Thanks @gaya3-zipstack.

src/unstract/sdk/index.py
```diff
@@ -72,44 +72,49 @@ def query_index(
...
-                self.tool.stream_log(f"Found {len(n.nodes)} nodes for {doc_id}")
-                all_text = ""
-                for node in n.nodes:
-                    all_text += node.get_content()
-                return all_text
-            else:
-                self.tool.stream_log(f"No nodes found for {doc_id}")
-                return None
+            try:
+                self.tool.stream_log(
```
gaya3-zipstack, line 76: Could move the try to after this log to make it more meaningful.

gaya3-zipstack (approved):
Added one minor suggestion. Otherwise looks good

src/unstract/sdk/index.py
```diff
@@ -73,43 +73,48 @@ def query_index(
...
+                else:
+                    self.tool.stream_log(f"No nodes found for {doc_id}")
+                    return None
+            except Exception as e:
+                self.tool.stream_log(
+                    f"Error while executing vector DB query: {e}", level=LogLevel.ERROR
+                )
+                raise VectorDBError(
+                    f"Failed to execute query on {vector_db}: {e}", actual_err=e
+                )
```
chandrasekharan-zipstack, line 117: NIT: For better readability and to avoid these nested try-excepts, we could move these into their own functions - `_build_query()` and `_execute_query()` maybe

chandrasekharan-zipstack (approved):
LGTM - left a minor comment to improve readability

src/unstract/sdk/index.py
pk-zipstack: Moving to separate functions would definitely improve readability. However, this bug is currently blocking core adapter functionality (Qdrant and Weaviate), so it's better we merge this fix first and we can create a follow-up task to refactor the code into `_build_query()` and `_execute_query()` functions.

PR #141, opened 2025-01-07
[FIX] Support for text extraction independent for indexing status.

harini-venkataraman:
## What
PR to support the extraction, even if the indexing is already done for the same document.
...

## Why
Since the doc_id is independent of tool_id, if we upload same document in a different project, extraction is not being done. This is because the doc_id matches.
...

## How
The fix is a workaround, changing the method invocation order. 
To be reconsidered for a permanent logical fix.
...

## Relevant Docs
Not applicable.
...

## Related Issues or PRs
Not applicable.
...

Not applicable.
...

## Notes on Testing
1. Upload a document in a project and index the document.
2. Upload the same document in a different project ( or maybe a newly created project ) and again index the document.
3. The raw view should be seen and to verify at dev perspective, the .txt file ( extracted file ) is to be present in the /user_id/tool_id/extract location.
...

## Screenshots

image

image

...

## Checklist


gaya3-zipstack (approved):
Looks good.

PR #145, opened 2025-01-15
passing file execution id from backend to tools

muhammad-ali-e:
## What

- Update the execution metadata with file_execution id

### Note:

- The transition from `run_id` to `file_execution_id` will be implemented incrementally. In some instances, the term `run_id` will be retained to minimize confusion during the review process due to the large number of touchpoints.

## Why

- Need to be added to get it in tools

## How

- By passing it from backend 

## Relevant Docs

-

## Related Issues or PRs

- https://github.com/Zipstack/unstract/pull/1065

-

## Notes on Testing

...

## Screenshots

...

## Checklist

### DIFF
--- a/src/unstract/sdk/constants.py
+++ b/src/unstract/sdk/constants.py
@@ -127,6 +127,7 @@ class MetadataKey:
     SOURCE_HASH = "source_hash"
     WORKFLOW_ID = "workflow_id"
     EXECUTION_ID = "execution_id"
+    FILE_EXECUTION_ID = "file_execution_id"
     ORG_ID = "organization_id"
     TOOL_META = "tool_metadata"
     TOOL_NAME = "tool_name"
@@ -168,3 +169,10 @@ class MimeType:
     PDF = "application/pdf"
     TEXT = "text/plain"
     JSON = "application/json"
+
+
+class UsageKwargs:
+    RUN_ID = "run_id"
+    FILE_NAME = "file_name"
+    WORKFLOW_ID = "workflow_id"
+    EXECUTION_ID = "execution_id"
--- a/src/unstract/sdk/tool/base.py
+++ b/src/unstract/sdk/tool/base.py
@@ -86,12 +86,11 @@ class BaseTool(ABC, StreamMixin):
         if parsed_args.command not in Command.static_commands():
             tool._exec_metadata = tool._get_exec_metadata()
             tool.workflow_id = tool._exec_metadata.get(MetadataKey.WORKFLOW_ID)
-            tool.execution_id = tool._exec_metadata.get(MetadataKey.EXECUTION_ID)
-            tool.execution_id = tool._exec_metadata.get(MetadataKey.EXECUTION_ID)
+            tool.execution_id = tool._exec_metadata.get(MetadataKey.EXECUTION_ID, "")
             tool.file_execution_id = tool._exec_metadata.get(
-                MetadataKey.FILE_EXECUTION_ID
+                MetadataKey.FILE_EXECUTION_ID, ""
             )
-            tool.source_file_name = tool._exec_metadata.get(MetadataKey.SOURCE_NAME)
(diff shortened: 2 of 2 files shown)


src/unstract/sdk/constants.py
```diff
@@ -168,3 +169,10 @@ class MimeType:
     PDF = "application/pdf"
     TEXT = "text/plain"
     JSON = "application/json"
+
+
+class UsageKwargs:
```
chandrasekharan-zipstack, line 174: @muhammad-ali-e where is this class of constants used? Did you define it for use in existing tools outside the SDK?

chandrasekharan-zipstack (approved):

src/unstract/sdk/constants.py
muhammad-ali-e: Yes. you can see the changes in https://github.com/Zipstack/unstract/pull/1065

pk-zipstack (approved):

src/unstract/sdk/tool/base.py
```diff
@@ -85,6 +87,11 @@ def from_tool_args(cls, args: list[str]) -> "BaseTool":
             tool._exec_metadata = tool._get_exec_metadata()
             tool.workflow_id = tool._exec_metadata.get(MetadataKey.WORKFLOW_ID)
             tool.execution_id = tool._exec_metadata.get(MetadataKey.EXECUTION_ID)
+            tool.execution_id = tool._exec_metadata.get(MetadataKey.EXECUTION_ID)
```
chandrasekharan-zipstack, line 90:
```suggestion
```
Repeated line

src/unstract/sdk/tool/base.py
```diff
@@ -85,6 +87,11 @@ def from_tool_args(cls, args: list[str]) -> "BaseTool":
...
+            tool.file_execution_id = tool._exec_metadata.get(
+                MetadataKey.FILE_EXECUTION_ID
+            )
+            tool.source_file_name = tool._exec_metadata.get(MetadataKey.SOURCE_NAME)
```
chandrasekharan-zipstack, line 94: @muhammad-ali-e would it be better to initialize this with `""` rather than defaulting to `None`? Since that's the default defined at the tool's level. I think we should propagate the same to `workflow_id` and `execution_id`

src/unstract/sdk/tool/base.py
```diff
@@ -85,6 +87,11 @@ def from_tool_args(cls, args: list[str]) -> "BaseTool":
             tool._exec_metadata = tool._get_exec_metadata()
             tool.workflow_id = tool._exec_metadata.get(MetadataKey.WORKFLOW_ID)
             tool.execution_id = tool._exec_metadata.get(MetadataKey.EXECUTION_ID)
+            tool.execution_id = tool._exec_metadata.get(MetadataKey.EXECUTION_ID)
+            tool.file_execution_id = tool._exec_metadata.get(
```
gaya3-zipstack, line 91: These fields get defaulted to None when not present. However these are initialised to "" in lines 42, 43. We should probably maintain the same defaults when the Key is not present also to avoid Type errors later on.

src/unstract/sdk/tool/base.py
muhammad-ali-e: Addressed but I feel we required some kind of validation later for these feilds if it empty or None.

src/unstract/sdk/tool/base.py
muhammad-ali-e: Addressed with default value,  but I feel we required some kind of validation later for these feilds if it empty or None.

src/unstract/sdk/tool/base.py
muhammad-ali-e: Addressed with default value, but I feel we required some kind of validation later for these feilds if it empty or None.

chandrasekharan-zipstack (approved):

PR #157, opened 2025-01-31
Remove feature flag related code changes for remote storage

gaya3-zipstack:
## What

1. Remove code bifurcation logic added for remote storage
2. Fix a small piece in summarisation
3. Allow MINIO as a Permanent File Storage Provider

## Why

1. To take remote storage to production
2. Prompt run with summarization as context gives an error saying "None does not have attribute startsWith" when we call exists() with a file that is None. 
3. Allow MINIO as permanent file storage provider to support OSS users.

## How

Manually refactor code for remote storage

## Relevant Docs

-

## Related Issues or PRs

-

-

## Notes on Testing

...

## Screenshots

...

## Checklist

### DIFF
--- a/src/unstract/sdk/file_storage/impl.py
+++ b/src/unstract/sdk/file_storage/impl.py
@@ -41,7 +41,7 @@ class FileStorage(FileStorageInterface):
 
         Args:
             path (str): Path to the file to be opened
-            mode (str): Mode in whicg the file is to be opened. Usual options
+            mode (str): Mode in which the file is to be opened. Usual options
                         include r, rb, w and wb
             encoding (str): Encoding type like utf-8 or utf-16
             seek_position (int): Position to start reading from
@@ -212,7 +212,6 @@ class FileStorage(FileStorageInterface):
             file_mtime = datetime.fromtimestamp(file_mtime)
         return file_mtime
 
-    @skip_local_cache
     def mime_type(
         self,
         path: str,
@@ -364,7 +363,6 @@ class FileStorage(FileStorageInterface):
             data: dict[str, Any] = yaml.safe_load(f)
             return data
 
-    @skip_local_cache
     def guess_extension(self, path: str) -> str:
         """Returns the extension of the file passed.
 
--- a/src/unstract/sdk/file_storage/impl.py
+++ b/src/unstract/sdk/file_storage/impl.py
@@ -216,7 +216,7 @@ class FileStorage(FileStorageInterface):
     def mime_type(
         self,
         path: str,
-        read_length: int = FileOperationParams.MIME_TYPE_DEFAULT_READ_LENGTH,
+        read_length: int = FileOperationParams.READ_ENTIRE_LENGTH,
     ) -> str:
         """Gets the file MIME type for an input file. Uses libmagic to perform
         the same.
@@ -224,7 +224,7 @@ class FileStorage(FileStorageInterface):
         Args:
             path (str): Path of the input file
             read_length (int): Length(bytes) to be read from the file for in
-            order to identify the mime type
+            order to identify the mime type. Defaults to read the entire length.
 
         Returns:
             str: MIME type of the file
--- a/src/unstract/sdk/file_storage/interface.py
+++ b/src/unstract/sdk/file_storage/interface.py
@@ -77,7 +77,7 @@ class FileStorageInterface(ABC):
     def mime_type(
         self,
         path: str,
-        read_length: int = FileOperationParams.MIME_TYPE_DEFAULT_READ_LENGTH,
+        read_length: int = FileOperationParams.READ_ENTIRE_LENGTH,
     ) -> str:
         pass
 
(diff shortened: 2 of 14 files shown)


Deepak-Kesavan (approved):

src/unstract/sdk/file_storage/impl.py
```diff
@@ -364,7 +363,6 @@ def yaml_load(
             data: dict[str, Any] = yaml.safe_load(f)
             return data
 
-    @skip_local_cache
```
ritwik-g, line 367: @gaya3-zipstack is it okay to remove this?

src/unstract/sdk/file_storage/permanent.py
```diff
@@ -27,7 +31,11 @@ def __init__(
                 f"supported in Permanent mode. "
                 f"Supported providers: {self.SUPPORTED_FILE_STORAGE_TYPES}"
             )
-        if provider == FileStorageProvider.GCS or provider == FileStorageProvider.LOCAL:
+        if (
+            provider == FileStorageProvider.GCS
+            or provider == FileStorageProvider.LOCAL
+            or provider == FileStorageProvider.MINIO
+        ):
```
muhammad-ali-e, line 38: NIT: I believe it would be great if it makes a list with a suitable category and use ` provider in` . this approach makes it feel like the provider is part of that category. 
Ex: `if provider in SUPPORTED_PROVIDERS:`

ritwik-g, line 38: Why this specific check is done? I think we already have a list of implemented Providers right? shouldn't we check that instead?

muhammad-ali-e (approved):
LGTM, left a minor comment

src/unstract/sdk/file_storage/permanent.py
ritwik-g: Also is this check even necessary?

src/unstract/sdk/index.py
```diff
@@ -297,8 +297,7 @@ def index(
 
             if doc_id_found and not reindex:
                 self.tool.stream_log(f"File was indexed already under {doc_id}")
-
-                if not fs.exists(output_file_path):
+                if output_file_path and not fs.exists(output_file_path):
```
ritwik-g, line 300: Could you add a comment in which case `output_file_path` will be None and what will happen in such a case?

src/unstract/sdk/file_storage/impl.py
gaya3-zipstack: Yes. Because the local cahe clear is already taken care in the read() function which is getting called in this function and also in . So if there is a problem, it will handled in read() itself. Hence this is redundant.

src/unstract/sdk/file_storage/permanent.py
gaya3-zipstack: We do not have specific implementation classes for each of the providers. If we do not have a chack here, how else to communicate to user that Azure for example is not supported(tested) yet?

src/unstract/sdk/index.py
gaya3-zipstack: There are cases where the file path is sent as None from Prompt-studio. @harini-venkataraman  or @Deepak-Kesavan will know better

src/unstract/sdk/file_storage/permanent.py
gaya3-zipstack: Will take of it in the next go @muhammad-ali-e Thank you. Your suggestion is definitely useful and a miss.

harini-venkataraman (approved):

src/unstract/sdk/file_storage/interface.py
```diff
@@ -77,7 +77,7 @@ def modification_time(self, path: str) -> datetime:
     def mime_type(
```
ritwik-g, line 77: @gaya3-zipstack an improvement suggestion. Looks like the mime_type detection might be same logic for any implementation. In such a case we can implement the logic here making this a normal method instead of static method. That way code duplication can be avoided

src/unstract/sdk/tool/base.py
```diff
@@ -127,34 +126,14 @@ def handle_static_command(self, command: str) -> None:
...
-                self.stream_error_and_exit(
-                    f"{file_to_get} is missing in EXECUTION_DATA_DIR"
-                )
-        else:
-            base_path: Path = self._get_data_dir()
-            file_path = base_path / file_to_get
-            if raise_err and not file_path.exists():
-                self.stream_error_and_exit(f"{file_to_get} is missing in TOOL_DATA_DIR")
+        base_path = self.execution_dir
+        file_path = base_path / file_to_get
```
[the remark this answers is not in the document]
ritwik-g: If yes this seems to raise error.  Also for file operation it might be better idea to os.path.join for path manipulations

image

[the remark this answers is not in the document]
muhammad-ali-e: @ritwik-g base_path might be a Path type. The Path type wil allow this.

PR #159, opened 2025-02-06
[FIX] Optional dependency groups for remote storage

harini-venkataraman:
## What
PR adds optional dependency groups for remote storage and fixes for Highlight issue
...

## Why
This ensure the package is not bloated, and user can refer to needed library downstream
...

## How
Added it as optional dependency in groups.
...

## Relevant Docs

-

## Related Issues or PRs

-

-

## Notes on Testing
Tested E2E flow, Indexing & prompt execution.
...

## Screenshots
Depedency resolved in backend
image

...

## Checklist


pyproject.toml
```diff
@@ -77,6 +77,10 @@ classifiers = [
 ]
 scripts = { unstract-tool-gen = "unstract.sdk.scripts.tool_gen:main" }
 
+[project.optional-dependencies]
+aws = ["s3fs[boto3]~=2024.10.0", "botocore~=1.34.162"]
```
gaya3-zipstack, line 81: Should we pin boto3 or botocore here? I think boto3 is the main one that drives botocore version.

harini-venkataraman: @gaya3-zipstack Agreed that boto3 and botocore have a parent-child relationship. However, for this use case, only botocore is required for s3fs. Including boto3 would introduce additional dependencies, leading to redundancy. [Ref](https://deps.dev/pypi/boto3/1.34.36/dependencies)

gaya3-zipstack: @harini-venkataraman  In this PR, we are still installing boto3 of s3fs. However we are explicitly pinning botocore and not boto3. My proposal is to check if we have to explicitly pin boto3 instead of botocore

gaya3-zipstack: Like:
aws = ["s3fs[boto3]~=2024.10.0", "boto3~=1.34.162"]

harini-venkataraman: @gaya3-zipstack Understood, like we discussed my concern was also overriding the dependency of boto3 pinned in s3fs. But yes we cannot skip the part. Making the suggested change.

pyproject.toml
```diff
@@ -77,6 +77,10 @@ classifiers = [
 ]
 scripts = { unstract-tool-gen = "unstract.sdk.scripts.tool_gen:main" }
 
+[project.optional-dependencies]
+aws = ["s3fs[boto3]~=2024.10.0", "boto3~=1.34.162"]
```
chandrasekharan-zipstack, line 81: @harini-venkataraman is it necessary to explicitly specify `boto3`? I don't see it being used directly and if its a dependent library - we shouldn't mention it here and let `s3fs[boto3]` bring whatever version it requires.
This can lead to resolution conflicts / long running resolution

src/unstract/sdk/adapters/x2text/llm_whisperer_v2/src/helper.py
```diff
@@ -369,7 +369,7 @@ def write_output_to_file(
             text_output = output_json.get("result_text", "")
             logger.info(f"Writing output to {output_file_path}")
             fs.write(
-                path=output_file_path, mode="w", data=text_output, encoding="utf-8"
+                path=str(output_file_path), mode="w", data=text_output, encoding="utf-8"
```
chandrasekharan-zipstack, line 372: NIT: Explicitly casting to `str` might not be required since `pathlib.Path` implements `__str__`. Does this cause an error?
There might be some caveats - for instance if explicitly check for `isinstance(file_path, str)` or use `str` specific functions not supported by `Path`

pyproject.toml
harini-venkataraman: @chandrasekharan-zipstack The boto3 version is pinned because s3fs is only compatible with botocore 1.34.x, and we encountered issues when other versions were referenced. To address this, we decided to pin boto3, the parent library of botocore.

To eliminate version conflicts, we performed version resolution across OSS and cloud plugins, determining the most compatible version.
cc @gaya3-zipstack

src/unstract/sdk/adapters/x2text/llm_whisperer_v2/src/helper.py
harini-venkataraman: @chandrasekharan-zipstack Yes, faced an error without explicit type casting.

gaya3-zipstack: Yeah, I was assuming the same @chandrasekharan-zipstack 
Infact simple string functions like print() works with Path. However, since we are passing these to typed functions, I am not sure if that can cause any trouble..

gaya3-zipstack (approved):
Looks good.

src/unstract/sdk/adapters/x2text/llm_whisperer_v2/src/helper.py
chandrasekharan-zipstack: @harini-venkataraman @gaya3-zipstack in that case I hope we've scoped our calls of these `FileStorage` methods. A clever solution might be to change the argument type to `Union[str, Path]` and within the lib method we can check for 
```python
if isinstance(file_path, Path):
    file_path = str(file_path)
```

This way we needn't tweak every call of the APIs

chandrasekharan-zipstack (approved):

src/unstract/sdk/adapters/x2text/llm_whisperer_v2/src/helper.py
harini-venkataraman: @chandrasekharan-zipstack We will add the optimization suggested in upcoming iterations, making additions to base method.

pyproject.toml
harini-venkataraman: @chandrasekharan-zipstack The version mentioned will take effect only if we use [aws] extras in the client side. Otherwise this will not be included in the packaging. For our use case, we had build and resolved the version and checked its compatibility. Any additions or changes to this client side will be reflected during locking process itself. 
Will add a comment here, to enhance developer experience.

gaya3-zipstack (approved):
looks good

PR #164, opened 2025-02-11
Support credentials to be empty in local filestorage

gaya3-zipstack:
## What

Support credentials to be empty in local filestorage

## Why

When local filestorage is configured as  TEST_LOCAL_STORAGE='{"provider":"local"}' in the env, the existing helper function errors out. This behaviour needs to be fixed

## How

Fixing the default for credentials to be dict 

## Relevant Docs

-

## Related Issues or PRs

-

-

## Notes on Testing

Test cases added for the same.

## Screenshots

...

## Checklist

### DIFF
--- a/src/unstract/sdk/file_storage/env_helper.py
+++ b/src/unstract/sdk/file_storage/env_helper.py
@@ -36,7 +36,7 @@ class EnvHelper:
             provider = FileStorageProvider(
                 file_storage_creds[CredentialKeyword.PROVIDER]
             )
-            credentials = file_storage_creds.get(CredentialKeyword.CREDENTIALS, "{}")
+            credentials = file_storage_creds.get(CredentialKeyword.CREDENTIALS, {})
             if storage_type == StorageType.PERMANENT:
                 file_storage = PermanentFileStorage(provider=provider, **credentials)
             elif storage_type == StorageType.SHARED_TEMPORARY:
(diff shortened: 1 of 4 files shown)


pk-zipstack (approved):

src/unstract/sdk/file_storage/env_helper.py
```diff
@@ -36,7 +36,7 @@ def get_storage(storage_type: StorageType, env_name: str) -> FileStorage:
             provider = FileStorageProvider(
                 file_storage_creds[CredentialKeyword.PROVIDER]
             )
-            credentials = file_storage_creds.get(CredentialKeyword.CREDENTIALS, "{}")
+            credentials = file_storage_creds.get(CredentialKeyword.CREDENTIALS, {})
```
harini-venkataraman, line 39: @gaya3-zipstack It would be better to have it as string with an empty dict for fallback, else we might bump into error where it might expect a string, but we would have sent a dict.
Also, Is a default value needed here, or can we raise an exception if envs are not set? Sometimes, fallbacks are misleading and breaks when fs_instance invokes any method.

harini-venkataraman (approved):
Added a comment, LGTM otherwise

src/unstract/sdk/file_storage/env_helper.py
gaya3-zipstack: As discussed, having it as a string will break the type check of PermanentFileStorage and the like. So keeping it as dict.
At the moment, I do not see a need to pass credentials as empty for cases like local storage as it is not required. Hence not adding an exception for the same. However if "provider" is missing, we already are handling it as KeyError.

PR #168, opened 2025-02-24
[FEAT] Support for Azure and S3 for permanent file storage

harini-venkataraman:
## What
PR adds support to Azure and S3 providers 
...

## Why
Onboarding and old customers use this provider
...

## How
Added support in SDK to be used downstream
...

## Relevant Docs
Not applicable
...

## Related Issues or PRs
Not applicable
...

...

## Notes on Testing
Added test cases for this new providers
...

## Screenshots

...

## Checklist


pk-zipstack (approved):

pyproject.toml
```diff
@@ -90,7 +91,8 @@ test = [
     "pytest==8.3.3",
     "pytest-mock==3.14.0",
     "gcsfs==2024.10.0",
-    "s3fs==2024.10.0"
+    "s3fs==2024.10.0",
+    "adlfs==2024.7.0"
```
ritwik-g, line 95: @harini-venkataraman why are we pinning the version here while the version is not pinned ni case of optional dependencies? Shouldn't we keep the version syntax same as the other place?

src/unstract/sdk/__init__.py
```diff
@@ -1,4 +1,4 @@
-__version__ = "0.58.0"
+__version__ = "0.58.1"
```
ritwik-g, line 1: @harini-venkataraman I believe the minor version needs to be updates since we are adding new support and this is not a bug fix.

src/unstract/sdk/file_storage/permanent.py
```diff
@@ -35,6 +36,8 @@ def __init__(
             provider == FileStorageProvider.GCS
             or provider == FileStorageProvider.LOCAL
             or provider == FileStorageProvider.MINIO
+            or provider == FileStorageProvider.S3
+            or provider == FileStorageProvider.AZURE
```
ritwik-g, line 40: @harini-venkataraman looking at the PR only this and the dependencies added seems to be the change. Is that enough to get this working?

ritwik-g (approved):
Overall changes looks okay. I got few queries for information purposes. And a comment about version

src/unstract/sdk/file_storage/permanent.py
harini-venkataraman: Yes @ritwik-g , we pass the protocol name along with the credentials to fsspec. Internally fsspec handles it.

pyproject.toml
harini-venkataraman: no specific reason here to pin, ideally it should be same as deps. Made the change

Deepak-Kesavan (approved):

src/unstract/sdk/file_storage/permanent.py
ritwik-g: Super.

ritwik-g (approved):

PR #169, opened 2025-02-26
[FIX] Added a condition to avoid pushing usage twice for few LLM models

Deepak-Kesavan:
## What

- Added a condition to avoid pushing usage twice for few LLM models

## Why

- Because it shows wrong usage details

## How

- 

## Relevant Docs

-

## Related Issues or PRs

-

-

## Notes on Testing

...

## Screenshots

...

## Checklist

### DIFF
--- a/src/unstract/sdk/utils/usage_handler.py
+++ b/src/unstract/sdk/utils/usage_handler.py
@@ -87,6 +87,7 @@ class UsageHandler(StreamMixin, BaseCallbackHandler):
             event_type == CBEventType.LLM
             and event_type not in self.event_ends_to_ignore
             and payload is not None
+            and "completion" in payload
         ):
             model_name = self.llm_model.metadata.model_name
             # Need to push the data to via platform service
--- a/src/unstract/sdk/utils/usage_handler.py
+++ b/src/unstract/sdk/utils/usage_handler.py
@@ -2,6 +2,7 @@ from typing import Any, Optional
 
 from llama_index.core.callbacks import CBEventType, TokenCountingHandler
 from llama_index.core.callbacks.base_handler import BaseCallbackHandler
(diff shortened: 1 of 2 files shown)


src/unstract/sdk/utils/usage_handler.py
```diff
@@ -87,6 +87,7 @@ def on_event_end(
             event_type == CBEventType.LLM
             and event_type not in self.event_ends_to_ignore
             and payload is not None
+            and "completion" in payload
```
chandrasekharan-zipstack, line 90: @Deepak-Kesavan are we sure that this key would be in the dictionary for all our LLM models? Would be better to use a typed class if its possible to avoid errors

chandrasekharan-zipstack (approved):
LGTM - left a minor comment alone

src/unstract/sdk/utils/usage_handler.py
Deepak-Kesavan: @chandrasekharan-zipstack Whenever we call `complete` function, the completion instance is returned. If the model is a chat model, then it is internally converted to completion instance in llama-index. But the issue is, the callback is called twice in case of chat models.

In our case these LLMs doesn't support completion.

Ollama
Mistral
Anthropic

Now, with the above condition in place we won't be writing it in DB twice even if the callback is called again. The usage record is added in the DB only when we receive the completion instance in the callback.

chandrasekharan-zipstack (approved):

src/unstract/sdk/utils/usage_handler.py
chandrasekharan-zipstack: Got it. I suggest adding a relevant comment against the code itself to indicate why we're doing so if it isn't immediately obvious

vishnuszipstack (approved):

PR #171, opened 2025-03-06
[Feat] o1-mini support for OpenAI/Azure OpenAI LLM adapters

pk-zipstack:
## What

Update of llama-index open ai packages for o1-mini support

## Why

The o1-mini reasoning models were not supported with our current llama-index openai packages.

## How

- Updated `llama-index-llms-openai` to `0.3.25`
- Updated `llama-index-llms-azure-openai` to `0.3.1`

## Relevant Docs

-

## Related Issues or PRs

-

-

## Notes on Testing

- Tested the model o1-mini in prompt studio, workflow, and API Deployment with both `OpenAI` and `Azure OpenAI` Adapters.

## Screenshots

...

## Checklist


pyproject.toml
```diff
@@ -34,13 +34,13 @@ dependencies = [
...
     "llama-index-vector-stores-qdrant==0.4.2",
-    "llama-index-llms-openai==0.3.17",
+    "llama-index-llms-openai==0.3.25",
     "llama-index-llms-palm==0.3.0",
     "llama-index-llms-mistralai==0.3.1",
     "mistralai==1.2.5",
     "llama-index-llms-anyscale==0.3.0",
     "llama-index-llms-anthropic==0.6.3",
-    "llama-index-llms-azure-openai==0.3.0",
+    "llama-index-llms-azure-openai==0.3.1",
```
chandrasekharan-zipstack, line 43: @pk-zipstack @gaya3-zipstack should we update other llama index LLMs as well?
Do you think we should stop pinning these versions to allow such fixes and enhancements?

chandrasekharan-zipstack (approved):

pyproject.toml
gaya3-zipstack: Just wondering whether we should only pin minor versions going forward for such dependencies.... May be we should take it up as a separate PR as that will involve some more testing if at all we decode to do that. This needs to be done only if we are delayed with liteLLM integration @chandrasekharan-zipstack @pk-zipstack

gaya3-zipstack (commented):
Pls generate pdm.lock file @pk-zipstack

pyproject.toml
pk-zipstack: @chandrasekharan-zipstack, currently, we only update the packages when a new model is required or when a customer asks. Therefore, I think it's best to ask @gaya3-zipstack about this.

For the second part, I also think it'll be better if we stop pinning the versions.

gaya3-zipstack (approved):
Looks good.

pyproject.toml
gaya3-zipstack: I think this needs to be taken up full-fledge....

PR #172, opened 2025-03-06
[FIX] Refactored indexing, extraction and retrieval

harini-venkataraman:
## What

- Separated indexing and extraction into distinct APIs in prompt service
- Refactored the SDK to accommodate the new APIs.
- Subquestion Retrieval using the Llama Index query engine.
- Refactored the Retriever service.
- Skipped VectorDB usage for zero chunk size.
- Added integration tests for the Indexing and Extraction APIs.

...

## Why
Pre requisite for cell type merge
...

## How
Refactored the existing code
...

## Relevant Docs

-

## Related Issues or PRs
https://github.com/Zipstack/unstract/pull/1172
https://github.com/Zipstack/unstract/pull/1149
...

-

## Notes on Testing
Added integration tests
...

## Screenshots

...

## Checklist


src/unstract/sdk/index_v2.py
```diff
@@ -0,0 +1,244 @@
...
+            )
+            self.tool.stream_log("File has been indexed successfully")
+        except Exception as e:
+            self.tool.stream_log(
+                f"Error adding nodes to vector db: {e}",
+                level=LogLevel.ERROR,
+            )
+            raise IndexingError(str(e)) from e
+
+    def _delete_existing_nodes_on_reindex(self, vector_db, doc_id, doc_id_found):
```
gaya3-zipstack, line 196: The function can be renamed to delete_nodes. That way it can be reused elsewhere if there is a need to delete nodes...

src/unstract/sdk/index_v2.py
```diff
@@ -0,0 +1,244 @@
...
+                    doc_id=doc_id,
+                    metadata={"section": item["section"]},
+                )
+                document.id_ = doc_id
+                documents.append(document)
+            self.tool.stream_log(f"Number of documents: {len(documents)}")
+            return documents
+        except Exception as e:
+            self.tool.stream_log(
+                f"Error deleting nodes for {doc_id}: {e}",
```
gaya3-zipstack, line 225: Wrong error msg.

src/unstract/sdk/extract.py
```diff
@@ -0,0 +1,20 @@
...
+        run_id: Optional[str] = None,
+        capture_metrics: bool = False,
+    ):
+        # TODO: Inherit from StreamMixin and avoid using BaseTool
+        self.tool = tool
+        self._run_id = run_id
+        self._capture_metrics = capture_metrics
+        self._metrics = {}
+
+    def extract(self):
```
gaya3-zipstack, line 19: where are we using this call?

src/unstract/sdk/exceptions.py
```diff
@@ -97,3 +97,12 @@ class FileOperationError(SdkError):
         "Please check specific storage error for "
         "further information"
     )
+
+
+class RetrievalError(SdkError):
+    """Custom exception raised for errors during retrieval from VectorDB."""
+
+    DEFAULT_MESSAGE = (
```
gaya3-zipstack, line 105: If this message is going all the way back to the user, then I think we need to shape this up a bit. User may not connect what the query param here would be. Instead we should rephrase it with parameters that user has control on.

src/unstract/sdk/index_v2.py
```diff
@@ -0,0 +1,244 @@
...
+        except Exception as e:
+            self.tool.stream_log(
+                f"Unexpected error during indexing check: {e}", level=LogLevel.ERROR
+            )
+
+        return doc_id_found
+
+    @log_elapsed(operation="INDEX")
+    @capture_metrics
+    def perform_indexing(
```
gaya3-zipstack, line 155: Thinking if we should retain old function names like index or index_document. That will keep the familiarity level intact for people who read the code.

gaya3-zipstack: Also applicable in other places  where relevant.

src/unstract/sdk/retrieval/simple.py
```diff
@@ -0,0 +1,53 @@
...
+
+        if not context:
+            # UN-1288 For Pinecone, we are seeing an inconsistent case where
+            # query with doc_id fails even though indexing just happened.
+            # This causes the following retrieve to return no text.
+            # To rule out any lag on the Pinecone vector DB write,
+            # the following sleep is added
+            # Note: This will not fix the issue. Since this issue is inconsistent
+            # and not reproducible easily, this is just a safety net.
+            time.sleep(2)
```
chandrasekharan-zipstack, line 51: NIT: This should only apply for pinecone then - why do we do this for all vectorDBs?
cc: @gaya3-zipstack

src/unstract/sdk/index_v2.py
```diff
@@ -0,0 +1,244 @@
...
+                return doc_id_found
+
+        except Exception as e:
+            self.tool.stream_log(
+                f"Unexpected error during indexing check: {e}", level=LogLevel.ERROR
+            )
+
+        return doc_id_found
+
+    @log_elapsed(operation="INDEX")
```
chandrasekharan-zipstack, line 153:
```suggestion
```
NIT: Remove this since you've already added it for the index API

src/unstract/sdk/index_v2.py
```diff
@@ -0,0 +1,244 @@
...
+        self.tool.stream_log("Indexing file...")
+        full_text = [
+            {
+                "section": "full",
+                "text_contents": extracted_text,
+            }
+        ]
+        # Convert raw text to llama index usage Document
+        documents = self._prepare_documents(doc_id, full_text)
+        self._delete_existing_nodes_on_reindex(vector_db, doc_id, doc_id_found)
```
chandrasekharan-zipstack, line 175: Why do we accept `doc_id_found` in this function? I feel that we should get rid of it and the caller of this function should ensure that we skip calling this function itself.

src/unstract/sdk/prompt.py
```diff
@@ -46,6 +46,32 @@ def answer_prompt(
...
+        self, payload: dict[str, Any], params: Optional[dict[str, str]] = None
+    ) -> dict[str, Any]:
+        url_path = "index"
+        if self.is_public_call:
+            url_path = "index-public"
+        return self._post_call(
+            url_path=url_path,
+            payload=payload,
+            params=params,
+        )
```
chandrasekharan-zipstack, line 61: @harini-venkataraman if we have an API to the `prompt-service` to take care of this, why do we have helper methods in the SDK? Shouldn't they be in prompt-service?

src/unstract/sdk/dto.py
```diff
@@ -0,0 +1,38 @@
+from dataclasses import dataclass, field
+from typing import Any, Optional
+
+
+@dataclass
+class InstanceIdentifiers:
+    embedding_instance_id: str
+    vector_db_instance_id: str
+    x2text_instance_id: str
+    llm_instance_id: str
```
chandrasekharan-zipstack, line 10: @harini-venkataraman I notice that we pass this object and `vector_db` / `embedding` adapter to functions. Can't we make use of the adapters alone everywhere? By making use of such UUIDs we force ourselves to use multiple DB queries every now and then

chandrasekharan-zipstack (commented):
Left some comments - my main question is on why we have such helper methods for indexing and an API call to the `prompt-service`

src/unstract/sdk/dto.py
```diff
@@ -0,0 +1,38 @@
...
+                "Indexing cannot be done for zero chunks."
+                "Please provide a valid chunk_size."
+            )
+
+
+@dataclass
+class ProcessingOptions:
+    reindex: bool = False
+    enable_highlight: bool = False
+    usage_kwargs: dict[Any, Any] = field(default_factory=dict)
```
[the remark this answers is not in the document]
harini-venkataraman: @gaya3-zipstack This will be passed to x2Text while initialising.

src/unstract/sdk/index_v2.py
harini-venkataraman: @gaya3-zipstack The parent API method is retained with the same name. Named the helper methods this way, so that the code is explanatory.

src/unstract/sdk/extract.py
```diff
@@ -0,0 +1,20 @@
...
+
+class Extract:
+    def __init__(
+        self,
+        tool: StreamMixin,
+        run_id: Optional[str] = None,
+        capture_metrics: bool = False,
+    ):
+        # TODO: Inherit from StreamMixin and avoid using BaseTool
+        self.tool = tool
```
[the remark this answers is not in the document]
harini-venkataraman: Extraction related logic is moved to prompt service, and this class is now removed.

src/unstract/sdk/index_v2.py
```diff
@@ -0,0 +1,244 @@
...
+                n: VectorStoreQueryResult = vector_db.query(query=q)
+                if len(n.nodes) > 0:
+                    doc_id_found = True
+                    self.tool.stream_log(f"Found {len(n.nodes)} nodes for {doc_id}")
+                else:
+                    self.tool.stream_log(f"No nodes found for {doc_id}")
+            except Exception as e:
+                self.tool.stream_log(
+                    f"Error querying {instance_identifiers.vector_db_instance_id}: {e},"
+                    " proceeding to index",
```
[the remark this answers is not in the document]
harini-venkataraman: @chandrasekharan-zipstack By "responsibility of the caller" you mean the indexing method should log this? 
Prefer to retain it this way, since we should not stop the indexing with querying of nodes fails for some reason.

src/unstract/sdk/index_v2.py
```diff
@@ -0,0 +1,244 @@
...
+            try:
+                n: VectorStoreQueryResult = vector_db.query(query=q)
+                if len(n.nodes) > 0:
+                    doc_id_found = True
+                    self.tool.stream_log(f"Found {len(n.nodes)} nodes for {doc_id}")
+                else:
+                    self.tool.stream_log(f"No nodes found for {doc_id}")
+            except Exception as e:
+                self.tool.stream_log(
+                    f"Error querying {instance_identifiers.vector_db_instance_id}: {e},"
```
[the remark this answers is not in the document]
harini-venkataraman: Logged the message instead of streaming it.

src/unstract/sdk/index_v2.py
```diff
@@ -0,0 +1,244 @@
...
+            )
+            raise SdkError(f"Error deleting nodes for {doc_id}: {e}") from e
+
+
+    def _is_no_op_adapter(
+        self,
+        instance_identifiers: InstanceIdentifiers,
+        vector_db: VectorDB,
+        doc_id: str,
+    ):
```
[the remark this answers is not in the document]
harini-venkataraman: The previous method signature returns `doc_id` and there might be cases where we use the value. Retaining the v1`s signature.

src/unstract/sdk/retrieval/simple.py
gaya3-zipstack: Because we do not have these vector DB methods in SDK overridden at the level of the individial adapter.  So we added it this way
CC: @chandrasekharan-zipstack

src/unstract/sdk/index_v2.py
```diff
@@ -0,0 +1,244 @@
...
+            # Delete the nodes for the doc_id
+            try:
+                vector_db.delete(ref_doc_id=doc_id)
+                self.tool.stream_log(f"Deleted nodes for {doc_id}")
+            except Exception as e:
+                self.tool.stream_log(
+                    f"Error deleting nodes for {doc_id}: {e}",
+                    level=LogLevel.ERROR,
+                )
+                raise SdkError(f"Error deleting nodes for {doc_id}: {e}") from e
```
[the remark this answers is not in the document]
harini-venkataraman: @chandrasekharan-zipstack Prefer to retain the exception handling here than the adapter, which varies based on the adapter.

src/unstract/sdk/exceptions.py
```diff
@@ -97,3 +97,12 @@ class FileOperationError(SdkError):
         "Please check specific storage error for "
         "further information"
     )
+
+
+class RetrievalError(SdkError):
```
chandrasekharan-zipstack, line 102: NIT: @harini-venkataraman I think that this can also be moved to the `prompt-service`. We don't seem to raise this anywhere in this PR. Ignore this comment if its already used somewhere in the SDK. I hope we aren't suppressing the actual error and returning this instead at such places

chandrasekharan-zipstack (approved):
LGTM - left a nit comment

src/unstract/sdk/prompt.py
harini-venkataraman: Addressed it. Thanks for suggesting.

src/unstract/sdk/exceptions.py
harini-venkataraman: Missed it while refactor. Thanks

gaya3-zipstack (approved):
Looks good.

src/unstract/sdk/__init__.py
```diff
@@ -1,4 +1,4 @@
-__version__ = "0.60.1"
+__version__ = "0.60.2"
```
gaya3-zipstack, line 1: Should we keep this 0.61.0 as we are introducing a couple of new APIs?

harini-venkataraman: I bumped patch version with an understanding that, there is not new functionalities introduced. But yeah, ultimately it is a new API. Bumping minor version.

gaya3-zipstack (approved):
looks good.

gaya3-zipstack (approved):
Looks fine.

PR #175, opened 2025-03-20
Feat: Support for Claude 3.7 sonnet LLM

pk-zipstack:
## What

- Support for Anthropic Claude 3.7 and thinking capabilities.

## Why

- Our current version of SDK does not support the latest claude 3.7 sonnet model.

## How

- Update the `llama-index-llms-anthropic` library.
- Added conditions to include the `thinking_dict` parameter for claude 3.7 model
- Updated the `json_schema.json` of anthropic to include a checkbox that enables the thinking feature of the model.

## Relevant Docs

-

## Related Issues or PRs

- https://zipstack.atlassian.net/browse/UN-2250

-

## Notes on Testing

- Tested connection of the anthropic LLM adapter with the claude 3.7 sonnet model with and without the thinking enabled
- Ran prompt studio, workflows, and API deployments with and without thinking capabilities with the LLM model.
- Tested with invalid budget_tokens when thinking is enabled i.e. budget tokens should be less than max_tokens and greater than 1024.
- Tested a scenario where thinking is enabled with models that don't support thinking. 

## Screenshots

image
image

- Scenario when thinking is enabled for models which don't support it
image

## Checklist


harini-venkataraman (approved):

Deepak-Kesavan:
@pk-zipstack Are we not planning to support thinking models?

cc: @gaya3-zipstack

pk-zipstack:
> @pk-zipstack Are we not planning to support thinking models?
> 
> cc: @gaya3-zipstack

@Deepak-Kesavan good point, I think we can have a check box to enable or disable thinking for the specific model.

gaya3-zipstack (commented):
@pk-zipstack We may have to support few additional params for this reasoning model to work well as mentioned by @arun-venkataswamy

pk-zipstack:
> @pk-zipstack We may have to support few additional params for this reasoning model to work well as mentioned by @arun-venkataswamy

Will look into it @gaya3-zipstack. Thanks.

Deepak-Kesavan (approved):
@pk-zipstack Looks good. I believe we can merge this @gaya3-zipstack . Later we can decide on displaying thinking response.

gaya3-zipstack (commented):
Does llama-index maintain a separate list for THINKING_MODELS? If yes, we may have to show the additonal config only on user entering those  models. However, our UI is controlled using a JSON schema today. Also, the models field is a text one. In order to support enabling/disabling the thinking params in UI based on the model entered by user, we may have to support additional APIs and Frontend should call those to decide whether these fields are to be shown. @Vishnu Sathyanesan  @Tahier Hussain @Deepak will know better

pk-zipstack:
> Does llama-index maintain a separate list for THINKING_MODELS? If yes, we may have to show the additonal config only on user entering those models. However, our UI is controlled using a JSON schema today. Also, the models field is a text one. In order to support enabling/disabling the thinking params in UI based on the model entered by user, we may have to support additional APIs and Frontend should call those to decide whether these fields are to be shown. @vishnu Sathyanesan @Tahier Hussain @deepak will know better

No @gaya3-zipstack, llama-index doesn't maintain a list like it is for the O series models for openai. I think this is because only one model i.e. 3.7-sonnet supports thinking. Even if the thinking is enabled for models which don't support them a notification is displayed saying the model doesn't support thinking. 

In future, if a list is added from the llama-index side for THINKING_MODELS, we can probably add an `OnChangeEvent` on the frontend to dynamically include the required fields.

gaya3-zipstack (approved):
Looks good.

gaya3-zipstack (approved):
Looks good.

PR #183, opened 2025-04-22
feat: Refactored platform and prompt helpers 

chandrasekharan-zipstack:
## What

- Refactor of helpers that connect to platform and prompt service
- Prompt service helper now raises an error on an exception instead of just logging and returning values
- Added provision to pass a `request_id` for tracing
- Bumped minor SDK version

## Why

- For better code reuse and maintainability

## Notes on Testing

- Tested locally

## Screenshots

image
## Checklist

### DIFF
--- a/src/unstract/sdk/__init__.py
+++ b/src/unstract/sdk/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "v0.70.1"
+__version__ = "v0.71.0"
 
 
 def get_sdk_version() -> str:
--- a/src/unstract/sdk/prompt.py
+++ b/src/unstract/sdk/prompt.py
@@ -2,6 +2,7 @@ import functools
 import logging
 from collections.abc import Callable
 from typing import Any, ParamSpec, TypeVar
+from deprecated import deprecated
 
 import requests
 from requests import ConnectionError, RequestException, Response
@@ -13,6 +14,7 @@ from unstract.sdk.constants import (
 from unstract.sdk.helper import SdkHelper
 from unstract.sdk.tool.base import BaseTool
 from unstract.sdk.utils.common_utils import log_elapsed
+from unstract.sdk.platform import PlatformHelper
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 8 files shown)


src/unstract/sdk/__init__.py
```diff
@@ -1,4 +1,4 @@
-__version__ = "v0.70.1"
+__version__ = "v0.71.0"
```
gaya3-zipstack, line 1: Why is this a minor version upgrade instead of patch? Is the request_id in Header a reason? @chandrasekharan-zipstack

chandrasekharan-zipstack: Thats one reason and the way we make use of the prompt / platform helper has changed. We basically raise an error directly from the helper now

gaya3-zipstack: Making use of helpers iin a different for should not affect any user if I understand it right. So that can still be accomodated as a PATCH version. But the request-id in header - is this a visible change for the user in any angle?

gaya3-zipstack: Even in tht case, only if the signature of any SDK API is changing, let us keep it a minor version. Else PATCH should be ok so that we stick to some consistency on the rules we had framed earlier.

chandrasekharan-zipstack: @gaya3-zipstack the function's interface has changed and will be a breaking change. For example, I've moved `get_exported_tool()` from `PromptTool` to `PlatformHelper`

> But the request-id in header - is this a visible change for the user in any angle?
Yes it is for the callers. Would allow for tracing requests

gaya3-zipstack: Ok. Hope backward compatibility is kept intact then... I mean a @deprecated decorator as we do normally and still support this for a few iterations @chandrasekharan-zipstack

gaya3-zipstack (approved):
Looks good.

src/unstract/sdk/__init__.py
chandrasekharan-zipstack: Added this

src/unstract/sdk/prompt.py
```diff
@@ -1,15 +1,63 @@
...
+            except RequestException as e:
+                error_message = str(e)
+                response = getattr(e, "response", None)
+                if response is not None:
+                    if (
+                        MimeType.JSON in response.headers.get("Content-Type", "").lower()
+                        and "error" in response.json()
+                    ):
+                        error_message = response.json()["error"]
+                    elif response.text:
```
harini-venkataraman, line 52: @chandrasekharan-zipstack Is a fallback needed here if response.text is None or empty of some reason?

src/unstract/sdk/prompt.py
```diff
@@ -1,15 +1,63 @@
...
+
+    def decorator(func: Callable[P, R]) -> Callable[P, R]:
+        @functools.wraps(func)
+        def wrapper(*args: P.args, **kwargs: P.kwargs) -> R:
+            try:
+                return func(*args, **kwargs)
+            except ConnectionError as e:
+                msg = f"Error while {context}. Unable to connect to prompt service."
+                logger.error(f"{msg}\n{e}")
+                args[0].tool.stream_error_and_exit(msg, e)
```
harini-venkataraman, line 42: @chandrasekharan-zipstack How is this assumption of `tool` in first element of args made? Does args have any one element?

harini-venkataraman (approved):
Left minor comments

src/unstract/sdk/prompt.py
chandrasekharan-zipstack: @harini-venkataraman we call the decorated function with `self` so that will be the first element always

src/unstract/sdk/prompt.py
chandrasekharan-zipstack: It will always be a response from a server. For the fallback, we use `str(e)` and set the error message just in case so it'll be taken care of

Deepak-Kesavan (approved):

PR #186, opened 2025-05-22
[FEAT] Llama-index LLM updates and reasoning effort parameter for OpenAI 'O' series models

pk-zipstack:
## What

- Updated Openai, vertexAI and Anthropic llama-index llm libraries.
- Added reasoning effort parameter for openai models

## Why

- customers requested O3 model for it's reasoning capabilities and low cost.
- Reasoning effort parameter gives control over the strength of reasoning for O series models.

## How

- Updated the respective libraries for the support of the new models.
- Included the parameter `reasoning_effort` from llama-index.

## Relevant Docs

-

## Related Issues or PRs

-

-

## Notes on Testing

Tested the updated models and the reasoning parameter by running the same in prompt-studio.

## Screenshots

image

## Checklist


src/unstract/sdk/__init__.py
```diff
@@ -1,4 +1,4 @@
-__version__ = "v0.71.1"
+__version__ = "v0.71.2"
```
chandrasekharan-zipstack, line 1:
```suggestion
__version__ = "v0.72.0"
```
Bump the minor version since its a new feature

src/unstract/sdk/adapters/llm/open_ai/src/static/json_schema.json
```diff
@@ -60,6 +60,53 @@
       "title": "Timeout",
       "default": 900,
       "description": "Timeout in seconds"
+    },
+    "enable_reasoning": {
+      "type": "boolean",
+      "title": "Enable Reasoning",
+      "default": false,
+      "description": "Allow the model to apply extra reasoning for complex tasks. May increase latency and cost. Only applicable for [O series models](https://platform.openai.com/docs/models#reasoning)."
```
chandrasekharan-zipstack, line 68: NIT: If we can quantify the increase in latency and cost it might be swell. However don't fret over this, an estimate based on your testing would also suffice

chandrasekharan-zipstack (approved):

pyproject.toml
```diff
@@ -26,7 +26,7 @@ dependencies = [
     "python-magic~=0.4.27",
     "python-dotenv==1.0.0",
     # Adapter changes
-    "llama-index==0.12.8",
+    "llama-index==0.12.37",
```
gaya3-zipstack, line 29: @pk-zipstack Are we not updating all adapter-packages alongwith this llama-index upgrade? Or the other llama-index packs have not gone past the mentioned versions here?
Eg. latest on  "llama-index-vector-stores-postgres" is still 0.4.1 only?

src/unstract/sdk/adapters/llm/open_ai/src/open_ai.py
```diff
@@ -70,10 +73,15 @@ def get_llm_instance(self) -> LLM:
                 "max_tokens": max_tokens,
             }
 
-            # O-series models default to temperature=1, ignoring passed values, so it's not set explicitly.
+            # O-series models default to temperature=1
             if model not in O1_MODELS:
                 llm_kwargs["temperature"] = 0
 
+            if enable_reasoning:
```
gaya3-zipstack, line 80: Should we have a check here to process reasoning related fields only if the models belong to O series?

pyproject.toml
pk-zipstack: @gaya3-zipstack I've only updated the LLMs packages to meet the specific model requirements, while the other packages were minor version changes. However, if you feel it's better to update all packages for consistency or other reasons, I'll proceed to make those changes and update the PR accordingly.

src/unstract/sdk/adapters/llm/open_ai/src/open_ai.py
pk-zipstack: It doesn't make sense, since llama-index does the same in their internal code. When the model is not of an "O" series, the reasoning effort doesn't get passed through. Even the enable reasoning parameter is redundant. I've added it so the user doesn't get confused when they don't want the reasoning feature. We can even remove the `enable reasoning` checkbox if that doesn't add anything to the UX.
CC: @chandrasekharan-zipstack

pyproject.toml
pk-zipstack: Vikash wanted the support for the o3 model for the testing of reasoning capabilities @gaya3-zipstack. I think it will be a blocker for him if we update all packages and perform testing across them since it will delay the support for the model. Can we merge this PR and update the all the other packages in a separate PR?

gaya3-zipstack: Let us proceed for now as you say this is a blocker
CC: @VikashPratheepan @pk-zipstack

gaya3-zipstack (approved):
Approving as this is a blocker for @VikashPratheepan 
Pls pick up version upgrade of adapters as soon as possible @pk-zipstack

PR #199, opened 2025-09-09
UN-2793 [FEAT] Added exponential backoff retry mechanism for platform service connections

chandrasekharan-zipstack:
## What

- Created generic retry decorator factory supporting multiple service prefixes with configurable exception types  
- Enhanced platform service calls with improved logging and structured backoff details
- Preserved existing sophisticated error detection logic for OSError errno codes
- Bumped SDK version to v0.78.0

## Why

- Platform service calls were failing immediately on connection errors without any retry attempts
- Better logging and observability for retry attempts and failures

## How

- **Created generic factory**: `create_retry_decorator` for flexible retry configurations
- **Environment configuration**: Prefix-based env vars allow different retry configs per service

## Can this PR break any existing features. If yes, please list possible items. If no, please explain why.

No, this PR should not break any existing features because:
- **Same error handling**: Errors are still raised after exhausting retries, preserving existing error handling  
- **Conservative defaults**: Default retry configuration remains the same (3 retries with exponential backoff)
- **Preserved error detection**: All existing OSError errno codes and HTTP status code logic maintained
- **Transparent to callers**: The retry mechanism maintains the same API and behavior

## Database Migrations

- N/A

**Environment variables (per service prefix, e.g., `PLATFORM_SERVICE`, `PROMPT_SERVICE`):**
- `{PREFIX}_MAX_RETRIES`: Maximum retry attempts (default: 3)
- `{PREFIX}_MAX_TIME`: Maximum total time in seconds (default: 60)  (please advice if this should be more @ritwik-g)
- `{PREFIX}_BASE_DELAY`: Initial delay in seconds (default: 1.0)
- `{PREFIX}_MULTIPLIER`: Backoff multiplier (default: 2.0)
- `{PREFIX}_JITTER`: Enable/disable jitter (default: true)

**Examples:**
- Platform service: `PLATFORM_SERVICE_MAX_RETRIES=5`
- Custom service: `API_SERVICE_MAX_RETRIES=2, API_SERVICE_BASE_DELAY=2.0`

## Relevant Docs

- N/A

## Related Issues or PRs

- [UN-2793](https://zipstack.atlassian.net/browse/UN-2793)

## Notes on Testing

- Tested locally by using it in the backend
- Didn't explicitly check in tools but it should work similarly

## Screenshots

image

## Checklist

I have read and understood the [Contribution Guidelines](https://docs.unstract.com/unstract/contributing/unstract/).

[UN-2793]: https://zipstack.atlassian.net/browse/UN-2793?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/unstract/sdk/utils/retry_utils.py
+++ b/src/unstract/sdk/utils/retry_utils.py
@@ -159,11 +159,10 @@ def retry_with_exponential_backoff(
                     if not should_retry or attempt == max_retries:
                         if attempt > 0:
                             logger_instance.exception(
-                                "Giving up '%s' after %d attempt(s) for %s: %s",
+                                "Giving up '%s' after %d attempt(s) for %s",
                                 func.__name__,
                                 attempt + 1,
                                 prefix,
-                                e,
                             )
                         raise
 


src/unstract/sdk/utils/retry_utils.py
```diff
@@ -0,0 +1,243 @@
...
+            # This should never be reached, but just in case
+            if last_exception:
+                raise last_exception
+
+        return wrapper
+
+    return decorator
+
+
+def retry_on_connection_error(
```
ritwik-g, line 209: @chandrasekharan-zipstack do we actually need this class? Couldn't we have used `retry_with_exponential_backoff` as that i also a decorator?

src/unstract/sdk/utils/retry_utils.py
```diff
@@ -0,0 +1,243 @@
...
+
+                except Exception as e:
+                    last_exception = e
+
+                    # Check if the error is retryable
+                    is_retryable = is_retryable_error(e)
+
+                    # Check additional retryable exceptions if provided
+                    if retryable_exceptions:
+                        is_retryable = is_retryable or isinstance(e, retryable_exceptions)
```
ritwik-g, line 175: @chandrasekharan-zipstack I feel like adding this here makes our implementation very tightly coupled and can't really be used generally. My suggestion would be keep `retry_with_exponential_backoff` generic so taht it can be used in any scenario.

May be we can pass a function like `skip_retry` which accepts an Exception and decides whether to retry for specific exception. That way this decorator won't be tightly coupled with underlying requirements.

In my mind it should be a very generic exponential retry decorator.

src/unstract/sdk/utils/retry_utils.py
```diff
@@ -0,0 +1,141 @@
...
+from collections.abc import Callable
+from typing import Any
+
+import backoff
+from requests.exceptions import ConnectionError, HTTPError, Timeout
+
+logger = logging.getLogger(__name__)
+
+
+def is_retryable_error(error: Exception) -> bool:
```
ritwik-g, line 15: @chandrasekharan-zipstack one of the specific error we got was err 111. Which is a TCP error. I don' think that is covered in this

src/unstract/sdk/utils/retry_utils.py
```diff
@@ -0,0 +1,141 @@
...
+        getattr(errno, "EHOSTUNREACH", 113),  # No route to host
+        getattr(errno, "ENETUNREACH", 101),  # Network is unreachable
+    }:
+        return True
+
+    return False
+
+
+def create_retry_decorator(
+    prefix: str = "PLATFORM_SERVICE",
```
ritwik-g, line 55: I would suggest not keeping this default value to avoid any future errors.

ritwik-g (commented):
@chandrasekharan-zipstack would be great if we can also cover prompt service retry with this.

gaya3-zipstack (approved):
Looks good to me.

src/unstract/sdk/utils/retry_utils.py
chandrasekharan-zipstack: Reproduced this locally and it seems to work fine if the 111 was raised as a `ConnectionError`
image

chandrasekharan-zipstack:
@ritwik-g I've addressed your comments and added support for prompt service as well

